PR #8, opened 2015-11-10
Don't remove file after releasing lock

kramarz:
Deleting file makes lock unreliable. File should exist in first place. Here is test that proves it.
http://s000.tinyupload.com/index.php?file_id=01315450548811350116
To run it you just need to install docker and execute ./run.sh as user from docker group.


AdorablePotato:
Thank you. This PR fixes also a problem with a test case :)

PR #16, opened 2017-01-09
lock.acquire() must be in the finally close

pypingou:
Otherwise the finally part is not being run I believe

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -98,7 +98,6 @@ Examples
     # instead:
     try:
         lock.acquire()
-        pass
     finally:
         lock.release()
 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -96,8 +96,8 @@ Examples
 
     # If you can not use the *with* statement, use a try-finally construct
     # instead:
-    lock.acquire()
     try:
+        lock.acquire()
         pass
     finally:
         lock.release()


docs/source/index.rst
```diff
@@ -96,8 +96,8 @@ Examples
 
     # If you can not use the *with* statement, use a try-finally construct
     # instead:
-    lock.acquire()
     try:
+        lock.acquire()
         pass
```
pombredanne: Would this `pass` still be needed then?

pypingou: Good point, it is likely not :)

AdorablePotato:
Hi, thank you. I think there is some ambiguousness in the examples.

The original example
```
lock.acquire()
try:
   pass
finally:
   lock.release()
```
was equivalent to:
```
with lock.acquire(...):
    pass
```

I'm not sure, but I think what you want with the changed version is something like
```
try:
   lock.acquire()
except TimeoutError:
   # Handle lock timeout.
   pass
else:
  try:
      # Code, that relies on the lock here.
      pass
  finally:
      lock.release()
```

I will change the order of the examples to make things clearer.

pypingou:
Hm ok, I think I see the idea, I would think it might be worth adding an ``# code that relies on the lock here and may throw exceptions`` above the ``pass`` command, as you did in the example you gave here :)
Just for clarity :)

PR #17, opened 2017-01-11
Logging

expobrain:
Added logging on acquiring and releasing a lock


AdorablePotato:
Thank you

moscoquera:
is the a way to toggle the log? it can add tons of unwanted log when you are doing tests on other apps.

expobrain:
@moscoquera you can setup the `filelock` logging to a level higher than INFO

PR #23, opened 2017-10-03
Change log level for failure to acquire lock to warn

Bezier89:


AdorablePotato:
I don't think there's a good reason to increase the log level. The user gets notified about a failed locking attempt by a Timeout exception and can then choose how to handle it (logging the exception including the traceback, trying again, ...). Changing the log level would also pollute the logs of some applications which rely heavily on filelocks. So better leave it up to the user if the Timeout exception should be logged or not.

PR #32, opened 2018-04-11
Corrected the prequel reference in README

KyleMSchmitt:
Just updated the quote.

👍 on having the correct spelling of Schmitt - "This is getting out of hand! Now there are two of them."

I hate it when he does that


AdorablePotato:
Thank you!

-- Sorry for the delay, I was on holiday :)

PR #42, opened 2018-11-01
Fix README rendering on PyPI

adamchainz:
The rST parser used on PyPI rejects anything with warnings, which is why the rendering is broken for the current version: https://pypi.org/project/filelock/3.0.9/

This can be checked with:

```
$ pip install docutils pygments
$ python setup.py check --metadata --restructuredtext
running check
warning: check: Duplicate implicit target name: "similar libraries".

warning: check: Duplicate implicit target name: "installation".

warning: check: Duplicate implicit target name: "documentation".

warning: check: Duplicate implicit target name: "contributions".

warning: check: Duplicate implicit target name: "license".
```

I fixed the warnings by removing the 'ToC' at the top. The README isn't so long so I don't think this is a huge loss. After this change, the check passes:

```
$ python setup.py check --metadata --restructuredtext
running check
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -3,12 +3,6 @@ py-filelock
 
 .. image:: https://travis-ci.org/benediktschmitt/py-filelock.svg?branch=master
     :target: https://travis-ci.org/benediktschmitt/py-filelock
-    
-`Similar Libraries <#similar-libraries>`_
-~ `Installation <#installation>`_
-~ `Documentation <#documentation>`_
-~ `Contributions <#contributions>`_
-~ `License <#license>`_
 
 This package contains a single module, which implements a platform independent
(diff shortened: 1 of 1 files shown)


AdorablePotato:
Thanks, I did not know about the check. Do you know, why it still looks that way?

adamchainz:
It won't be updated until it gets uploaded by a release :)

adamchainz:
Okay maybe you did do a release. I'm not sure, but the `license` field is being filled in with the full license text, when it's intended for just the name of the license (e.g. https://github.com/adamchainz/ec2-metadata/blob/master/setup.py#L41 ). Try removing that? You can use Test PyPI to create trial uploads rather than doing repeat releases to production PyPI.

PR #47, opened 2019-01-12
Handle Unwritable Path

ooglek:
## Description

If the OS cannot open a file because the path is bad, it does not make sense to repeatedly attempt to open the file as it will continue to fail, indefinitely if the `-1` default timeout is used, without any feedback to the user.

This modifies the behavior to raise the OSError/IOError exception received on Windows or Unix so FileLock exits rather than a futile infinite loop that will never succeed.

Tests are written to demonstrate the behavior.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,7 +28,7 @@ repos:
     rev: v1.11.0
     hooks:
       - id: blacken-docs
-        additional_dependencies: [ black==20.8b1 ]
+        additional_dependencies: [ black==21.9b0 ]
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.9.0
     hooks:
--- a/src/filelock/__init__.py
+++ b/src/filelock/__init__.py
@@ -334,7 +334,7 @@ class WindowsFileLock(BaseFileLock):
         try:
             fd = os.open(self._lock_file, open_mode)
         except OSError:
-            pass
+            raise
         else:
             try:
                 msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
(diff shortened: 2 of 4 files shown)


AdorablePotato:
Thank you!
I merged your pull request and made some changes to SoftFileLock: https://github.com/benediktschmitt/py-filelock/tree/ooglek-handle-unwritable-path

I can only run tests on linux at the moment, so I'd appreciate any feedback for other systems.

mcsimps2:
@benediktschmitt I don't see these changes in master and it still seems to be a problem in the latest pip release #60.  It looks like the changes are only in the branch `ooglek-handle-unwritable-path` and not in `master`.

hmeine:
I am very much interested in this fix. Indeed, we had two situations (after switching from a different lock implementation to filelock) in which SoftFileLock would just hang:
* a read-only filesystem
* a non-existant directory

In both cases, one would expect an exception to be raised that allows to discriminate from an existing lock file.

gaborbernat (changes requested):

ooglek:
I included a test for the change in the original pull request 2+ years ago. Hopefully it should suffice.

gaborbernat:
Oh missed that, still needs a rebase though 😊

ooglek:
@mcsimps2 @hmeine My need for this passed 2 years ago. One of you or someone you know will need to carry on this one-line change if you want it in a release.

FFY00:
I rebased the PR.

gaborbernat (changes requested):
We'll need to fix the tests and linting 👍

FFY00:
Yes, I had a look and the issues seem to be false positives (unless my english is rusty and there is actually a typo in words flagged).

gaborbernat:
Superseded by https://github.com/tox-dev/py-filelock/pull/96

PR #55, opened 2019-08-23
It is unsafe to do IO from __del__

hackaugusto:
This PR removes the logging call from the destructor. Since the time of the execution of a destructor is undefined, it can happen while the interpreter is being teared down, at this point there are no guarantees that any of the subsystems will be available to it.

Using the logging system at this point is not a good idea and can lead to very weird errors https://github.com/raiden-network/raiden/issues/4639

### DIFF
--- a/filelock.py
+++ b/filelock.py
@@ -290,7 +290,7 @@ class BaseFileLock(object):
             raise
         return _Acquire_ReturnProxy(lock = self)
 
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):

gaborbernat:
Closing due to lack of response, we'll reopen if this gets traction again.

PR #66, opened 2020-04-05
Avoid 1s delay on consequent lock in windows

awaizman1:
use winapi LockFile instead of msvcrt.locking in order to avoid 1s delay when file is locked

in windows when the file is locked (using msvcrt.locking) another attempt to lock it will take 1s (just to see its already locked). using winapi FileLock instead doesn't have this buggy implemetation.
for more details see [this issue](https://github.com/benediktschmitt/py-filelock/issues/65)

### DIFF
--- a/filelock.py
+++ b/filelock.py
@@ -41,6 +41,7 @@ except ImportError:
 
 try:
     import msvcrt
+    from ctypes import windll, wintypes, WinError
 except ImportError:
     msvcrt = None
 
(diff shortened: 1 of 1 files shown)


gaborbernat:
Can you please rebase it on the latest main? Thanks!

gaborbernat (commented):

gaborbernat (changes requested):

gaborbernat:
Closing due to lack of response, we'll reopen if this gets traction again.

PR #86, opened 2021-09-22
Tidy setup.py

inverse:
- Specify min Python version (3.6 as prior versions are EOL)
- Tidy description loading

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,16 +33,15 @@ from filelock import __version__
 
 # Main
 # ------------------------------------------------
-try:
-    long_description = open("README.md").read()
-except (OSError, IOError):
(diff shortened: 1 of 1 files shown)


FFY00 (approved):
It probably would make sense to move to `setup.cfg`.

inverse:
@FFY00 or adopting Poetry to ease dependency/packaging

FFY00:
It does not support editable installs yet.

inverse:
@FFY00 I get the need for such thing when developing a modular system but would that be a blocker for something like this that is more of a utility package?

gaborbernat:
This has been done in #87

PR #96, opened 2021-09-28
Raise when trying to acquire in R/O or missing folder

gaborbernat:
Signed-off-by: Bernát Gábor <gaborjbernat@gmail.com>

### DIFF
--- a/src/filelock/_soft.py
+++ b/src/filelock/_soft.py
@@ -1,15 +1,15 @@
 import os
-from errno import ENOENT
+from errno import EACCES, ENOENT
 
 from ._api import BaseFileLock
-from ._util import raise_on_ro_file
+from ._util import raise_on_exist_ro_file
 
 
 class SoftFileLock(BaseFileLock):
     """Simply watches the existence of the lock file."""
 
     def _acquire(self):
-        raise_on_ro_file(self._lock_file)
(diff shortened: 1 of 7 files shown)


src/filelock/_soft.py
```diff
@@ -10,8 +14,11 @@ def _acquire(self):
         open_mode = os.O_WRONLY | os.O_CREAT | os.O_EXCL | os.O_TRUNC
         try:
             fd = os.open(self._lock_file, open_mode)
-        except OSError:
-            pass
+        except OSError as exception:
+            if exception.errno in (EPERM, EACCES, ENOENT):
+                raise
+            if sys.platform == "win32" and not os.access(self._lock_file, os.W_OK):
+                raise PermissionError("Permission denied")
```
FFY00: Maybe `raise ... from None` to remove the exception chaining.

gaborbernat: that doesn't work on python 2, that we still support here; perhaps once we drop it

FFY00: Ah, of course.

gaborbernat:
I'm not sure we can make the soft file lock deadlock safe 🤦 but will keep trying 😟

gaborbernat (commented):
The end result feels hacky, but hey it works 🤷

PR #97, opened 2021-10-01
Changed logger name from "filelock._api" to "filelock"

### LINKED ISSUE #59: [Request] Add option to disable logger
By default this package will always print to a logger, but for my use case it only ends up cluttering the logs with "released, acquired, released, acquired, etc". I'd appreciate it if you could add the option to disable the logger :)

hkennyv:
Fixes the logger name and closes #59

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,18 @@ with lock:
 ```
 
 
+### Adjusting the Log Level
+
+By default, the logger name is *"filelock"* and the logging level is set to
+*logging.WARNING*. This can be adjusted using the *logging* module.
+
+```Python
+import logging
+
+logging.getLogger("filelock").setLevel(logging.ERROR)
+```
+
+
 ## FileLock vs SoftFileLock
 
 The *FileLock* is platform dependent while the *SoftFileLock* is not. Use the
--- a/README.md
+++ b/README.md
@@ -111,17 +111,12 @@ with lock:
 # And released here.
 ```
 
+All log messages by this library are made using the *DEBUG* level. On how to
+control displaying/hiding that please consult the
+[logging documentation of the standard library](https://docs.python.org/3/howto/logging.html).
 
-### Adjusting the Log Level
-
-By default, the logger name is *"filelock"* and the logging level is set to
-*logging.WARNING*. This can be adjusted using the *logging* module.
-
-```Python
-import logging
-
-logging.getLogger("filelock").setLevel(logging.ERROR)
-```
+E.g. to hide these messages you can use
+`logging.getLogger("filelock").setLevel(logging.INFO)`.
 
 
 ## FileLock vs SoftFileLock
--- a/README.md
+++ b/README.md
@@ -111,8 +111,8 @@ with lock:
 # And released here.
 ```
 
-All log messages by this library are made using the *DEBUG* level. On how to
-control displaying/hiding that please consult the
+All log messages by this library are made using the *DEBUG* level, under the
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -112,6 +112,18 @@ with lock:
 ```
 
 
+### Adjusting the Log Level
+
+By default, the logger name is *"filelock"* and the logging level is set to
+*logging.WARNING*. This can be adjusted using the *logging* module.
```
gaborbernat: This was INFO, but we changed it to a warning. Now all log messages are DEBUG AFAIK.

tests/test_filelock.py
```diff
@@ -333,3 +334,14 @@ def test_cleanup_soft_lock(tmp_path):
     with lock:
         assert lock_path.exists()
     assert not lock_path.exists()
+
+
+def test_logger_name():
+    # need to check both that the filelock logger exists and is not created as
```
gaborbernat: Please align with 120 characters width 👍 but IMHO anything this tests does is already tested above and therefore can be removed 👍

gaborbernat (changes requested):

tests/test_filelock.py
hkennyv: @gaborbernat Wow that speedy review. Sure, I wasn't sure if we wanted an explicit test to test the log name rather than grouping it with the `test_simple`. I can remove it 👍

README.md
hkennyv: @gaborbernat right, I meant to describe the logger's default level (since its not specified, it defaults to [`WARNING`](https://docs.python.org/3/howto/logging.html#when-to-use-logging)) rather than the level of the messages that the library emits.

Any suggestions on wordsmithing here to make that more clear? How about:

---
### Verbose Logging

Verbose logging can be enabled by setting the log level of the *"filelock"* logger to *logging.DEBUG*.

```Python
import logging

logging.getLogger("filelock").setLevel(logging.DEBUG)
```

gaborbernat: I don't think we need to document this, and your description actually is not true because things showing up depends on the level for both the logger and the log handler. IMHO we can just say that:

All log messages by this library are made using the DEBUG level. On how to control displaying/hiding that please consult the [logging documentation of the standard library](https://docs.python.org/3/howto/logging.html`). E.g. to hide these messages you can use `logging.getLogger("filelock").setLevel(logging.INFO)`.


We can't say anything about actually showing it because for that we'd need to know how the log handler is set up.

src/filelock/_api.py
```diff
@@ -4,7 +4,7 @@
 
 from ._error import Timeout
 
-_LOGGER = logging.getLogger(__name__)
+_LOGGER = logging.getLogger("filelock")
```
gaborbernat, line 7: I think we should add here:

```
_LOGGER.set_level(logging.DEBUG)
```

README.md
hkennyv: That's a good point. See the updated changes

gaborbernat (approved):

README.md
```diff
@@ -111,6 +111,13 @@ with lock:
 # And released here.
 ```
 
+All log messages by this library are made using the *DEBUG* level. On how to
```
gaborbernat: All log messages by this library are made using the *DEBUG* level, under the `filelock` name. 

👍 Forgot to document the name.

hkennyv: Done 🚀

hkennyv:
@gaborbernat updated & thanks for the reviews! LMK if there's anything else I need to do.

PR #110, opened 2021-10-22
Update `FileLock` constructor to accept `pathlib.Path`

f-koehler:
This is a small fix for people who employ `pathlib.Path` for their path handling. It allows a `Path` object to be passed to the constructor of `FileLock`. It is then converted to a normal `str` when storing it inside the object leaving the rest of the library unaffected.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -4,7 +4,7 @@ import time
 from abc import ABC, abstractmethod
 from threading import Lock
 from types import TracebackType
-from typing import Optional, Type, Union
+from typing import Any, Optional, Type, Union
 
 from ._error import Timeout
 
@@ -35,7 +35,7 @@ class AcquireReturnProxy:
 class BaseFileLock(ABC):
     """Abstract base class for a file lock object."""
 
-    def __init__(self, lock_file: Union[str, os.PathLike], timeout: float = -1) -> None:
+    def __init__(self, lock_file: Union[str, "os.PathLike[Any]"], timeout: float = -1) -> None:
         """
         Create a new lock object.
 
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
Test and changelog? 😊

f-koehler:
> Test and changelog? blush

I added a test using `pytest.mark.parametrize`. What are you referring to with "changelog" am I supposed to update some changelog file. Sorry if I am being stupid but I could not find any instructions in the repo/doc. Only the `docs/changelog.rst` file which I did not want to touch since you seem to be the only author of that.

So please let me know what's missing.

gaborbernat:
> Only the `docs/changelog.rst` file which I did not want to touch since you seem to be the only author of that.

That's the one. Ignore that in past only I've edited it.

FFY00 (changes requested):
Can we please make it `os.PathLike` instead? It covers way more types and the implementation is mostly the same, you just replace `str()` with `os.fspath()`.

f-koehler:
@FFY00 and @gaborbernat, I will get to work on this next week.

FFY00:
Sounds good, no worries :blush:

src/filelock/_api.py
```diff
@@ -44,7 +45,7 @@ def __init__(self, lock_file: str, timeout: float = -1) -> None:
          A timeout of 0 means, that there is exactly one attempt to acquire the file lock.
         """
         # The path to the lock file.
-        self._lock_file: str = lock_file
+        self._lock_file: str = os.fspath(lock_file)  # noqa: SC200
```
gaborbernat: Please remove this noqa and fix it by adding to to the whitelist.txt.

gaborbernat (changes requested):

FFY00 (approved):
This looks alright to me, thanks!

gaborbernat (approved):

gaborbernat:
https://pypi.org/project/filelock/3.3.2/ is now out

PR #113, opened 2021-10-27
Create lockfile on Linux with safe permissions

ghost:
According to the documentation of `os.open` the former `mode` was in fact the `flags` for the `open` operation. I renamed this from `open_mode` to `open_flags` and added the permission as `0o660` so that the created lockfile now has `-rw-rw----` and not `-rwxr-xr-x` as before.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,7 +3,7 @@ Changelog
 
 v3.3.2 (2021-10-28)
 -------------------
-- Ensure that newly created lockfiles have 660 permissions :pr:`113` - by :user:`kbojens`
+- Ensure that newly created lockfiles now have `-rw-rw----` permissions and not `-rwxr-xr-x` as before :pr:`113` - by :user:`kbojens`
 
 v3.3.1 (2021-10-15)
 -------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,7 +3,8 @@ Changelog
 
 v3.3.3 (2021-10-28)
 -------------------
-- Ensure that newly created lockfiles now have ``-rw-rw----`` permissions and not ``-rwxr-xr-x`` as before :pr:`113` - by :user:`kbojens`
+- Ensure that newly created lockfiles now have ``-rw-rw----`` permissions and not ``-rwxr-xr-x`` as before :pr:`113` -
+  by :user:`kbojens`
 
 v3.3.2 (2021-10-29)
 -------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,7 +3,7 @@ Changelog
 
 v3.3.2 (2021-10-28)
 -------------------
-- Ensure that newly created lockfiles now have `-rw-rw----` permissions and not `-rwxr-xr-x` as before :pr:`113` - by :user:`kbojens`
+- Ensure that newly created lockfiles now have ``-rw-rw----`` permissions and not ``-rwxr-xr-x`` as before :pr:`113` - by :user:`kbojens`
 
 v3.3.1 (2021-10-15)
 -------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,7 +3,7 @@ Changelog
 
 v3.3.3 (2021-10-28)
 -------------------
-- Ensure that newly created lockfiles now have `-rw-rw----` permissions and not `-rwxr-xr-x` as before :pr:`113` - by :user:`kbojens`
+- Ensure that newly created lockfiles now have ``-rw-rw----`` permissions and not ``-rwxr-xr-x`` as before :pr:`113` - by :user:`kbojens`
 
 v3.3.2 (2021-10-29)
 -------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,10 +1,15 @@
 Changelog
 =========
 
-v3.3.3 (2021-10-28)
+v3.5.9 (2021-12-04)
 -------------------
 - Ensure that newly created lockfiles now have ``-rw-rw----`` permissions and not ``-rwxr-xr-x`` as before :pr:`113` -
   by :user:`kbojens`
+
+v3.4.0 (2021-11-16)
+-------------------
+- Add correct spelling of poll interval parameter for :meth:`acquire <filelock.BaseFileLock.acquire>` method, raise
+  deprecation warning when using the misspelled form :pr:`119` - by :user:`XuehaiPan`.
 
 v3.3.2 (2021-10-29)
 -------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+v3.3.3 (2021-10-28)
+-------------------
+- Ensure that newly created lockfiles have 660 permissions :pr:`113` - by :user:`kbojens`
+
 v3.3.2 (2021-10-29)
 -------------------
 - Accept path types (like ``pathlib.Path`` and ``pathlib.PurePath``) in the constructor for ``FileLock`` objects.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+v3.3.2 (2021-10-28)
+-------------------
+- Ensure that newly created lockfiles have 660 permissions :pr:`113` - by :user:`kbojens`
+
 v3.3.1 (2021-10-15)
 -------------------
 - Add changelog to the documentation :pr:`108` - by :user:`gaborbernat`
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,9 +1,14 @@
 Changelog
 =========
 
-v3.3.2 (2021-10-28)
+v3.3.3 (2021-10-28)
 -------------------
-- Ensure that newly created lockfiles now have ``-rw-rw----`` permissions and not ``-rwxr-xr-x`` as before :pr:`113` - by :user:`kbojens`
+- Ensure that newly created lockfiles now have ``-rw-rw----`` permissions and not ``-rwxr-xr-x`` as before :pr:`113` -
+  by :user:`kbojens`
+
+v3.3.2 (2021-10-29)
+-------------------
+- Accept path types (like ``pathlib.Path`` and ``pathlib.PurePath``) in the constructor for ``FileLock`` objects.
 
 v3.3.1 (2021-10-15)
 -------------------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,7 +3,7 @@ Changelog
 
 v3.3.3 (2021-10-28)
 -------------------
-- Ensure that newly created lockfiles have 660 permissions :pr:`113` - by :user:`kbojens`
+- Ensure that newly created lockfiles now have `-rw-rw----` permissions and not `-rwxr-xr-x` as before :pr:`113` - by :user:`kbojens`
 
 v3.3.2 (2021-10-29)
 -------------------
--- a/src/filelock/_soft.py
+++ b/src/filelock/_soft.py
@@ -18,8 +18,9 @@ class SoftFileLock(BaseFileLock):
             | os.O_EXCL  # together with above raise EEXIST if the file specified by filename exists
             | os.O_TRUNC  # truncate the file to zero byte
         )
+        mode = 0o660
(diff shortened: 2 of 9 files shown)


gaborbernat (changes requested):
YOu'll need to add a test and a changelog entry 👍

ghost:
> YOu'll need to add a test and a changelog entry

Done. There is just a problem with the linting:

```
tests/test_filelock.py:7:20: SC200 Possibly misspelt word: 'IMODE'
tests/test_filelock.py:350:18: SC200 Possibly misspelt word: 'IMODE'
```

I was not sure where to whitelist this word.

jugmac00:
> Done. There is just a problem with the linting:
> 
> ```
> tests/test_filelock.py:7:20: SC200 Possibly misspelt word: 'IMODE'
> tests/test_filelock.py:350:18: SC200 Possibly misspelt word: 'IMODE'
> ```
> 
> I was not sure where to whitelist this word.

https://github.com/tox-dev/py-filelock/blob/main/whitelist.txt

ghost:
> > Done. There is just a problem with the linting:
> > ```
> > tests/test_filelock.py:7:20: SC200 Possibly misspelt word: 'IMODE'
> > tests/test_filelock.py:350:18: SC200 Possibly misspelt word: 'IMODE'
> > ```

> https://github.com/tox-dev/py-filelock/blob/main/whitelist.txt

Thanks. That worked. Although I don't understand why flake8 warns about `IMODE` when the actual import is `S_IMODE` …

jugmac00:
> > > Done. There is just a problem with the linting:
> > > ```
> > > tests/test_filelock.py:7:20: SC200 Possibly misspelt word: 'IMODE'
> > > tests/test_filelock.py:350:18: SC200 Possibly misspelt word: 'IMODE'
> > > ```
> 
> > https://github.com/tox-dev/py-filelock/blob/main/whitelist.txt
> 
> Thanks. That worked. Although I don't understand why flake8 warns about `IMODE` when the actual import is `S_IMODE` …

It is not flake8, but flake8-spellcheck

I guess the plugin uses some kind of splitting.

whitelist.txt
```diff
@@ -38,3 +38,4 @@ unlck
 util
 win32
 wronly
+IMODE
```
gaborbernat: this needs to be all lowercase as is the rest

tests/test_filelock.py
```diff
@@ -337,3 +338,15 @@ def test_cleanup_soft_lock(tmp_path: Path) -> None:
     with lock:
         assert lock_path.exists()
     assert not lock_path.exists()
+
+
+@pytest.mark.parametrize("lock_type", [FileLock])
+def test_file_permissions(lock_type: Type[BaseFileLock], tmp_path: Path) -> None:
```
gaborbernat: I don't think you need a new test you can extend existing ones 😊

src/filelock/_soft.py
```diff
@@ -12,14 +12,14 @@ class SoftFileLock(BaseFileLock):
...
-        mode = (
+        flags = (
             os.O_WRONLY  # open for writing only
             | os.O_CREAT
             | os.O_EXCL  # together with above raise EEXIST if the file specified by filename exists
             | os.O_TRUNC  # truncate the file to zero byte
         )
         try:
-            fd = os.open(self._lock_file, mode)
+            fd = os.open(self._lock_file, flags)
```
gaborbernat: any reason we don't need mode here?

docs/changelog.rst
```diff
@@ -1,6 +1,10 @@
 Changelog
 =========
 
+v3.3.2 (2021-10-28)
+-------------------
+- Ensure that newly created lockfiles have 660 permissions :pr:`113` - by :user:`kbojens`
```
gaborbernat: instead of the opt code can you use the human from instead (`rw-`...)?

gaborbernat (changes requested):

whitelist.txt
ghost: Done.

tests/test_filelock.py
ghost: I've now extended and verified the test_simple()

src/filelock/_soft.py
ghost: You are right. I was somehow under the impression that this would not create new files.

docs/changelog.rst
ghost: Done.

tests/test_filelock.py
```diff
@@ -34,6 +35,11 @@ def test_simple(lock_type: Type[BaseFileLock], tmp_path: Path, caplog: LogCaptur
     assert [r.name for r in caplog.records] == ["filelock", "filelock", "filelock", "filelock"]
     assert logging.getLogger("filelock").level == logging.NOTSET
 
+    lock.acquire()
```
gaborbernat: I think we can move the check to line 23, and not need the extra lock /acquire

gaborbernat (changes requested):

gaborbernat:
@kbojens any ideas why within the context manager is not 0660 but 0640?

ghost:
> @kbojens any ideas why within the context manager is not 0660 but 0640?

Well, that's interesting. On my local machine all the tests run without _any_ problem:

```
    with lock as locked:
        assert lock.is_locked
        assert lock is locked
        assert oct(S_IMODE(os.stat(lock_path).st_mode)) == oct(0o660)
    assert not lock.is_locked
```

```
============================= test session starts ==============================
collecting ... collected 33 items

tests/test_filelock.py::test_simple[UnixFileLock-str] PASSED             [  3%]
tests/test_filelock.py::test_simple[UnixFileLock-PurePath] PASSED        [  6%]
tests/test_filelock.py::test_simple[UnixFileLock-Path] PASSED            [  9%]
tests/test_filelock.py::test_simple[SoftFileLock-str] PASSED             [ 12%]
tests/test_filelock.py::test_simple[SoftFileLock-PurePath] PASSED        [ 15%]
tests/test_filelock.py::test_simple[SoftFileLock-Path] PASSED            [ 18%]

```

```
18:36:54 fedora /tmp/pytest-of-kbojens/pytest-1 $> ls -al test_simple_*/a
-rw-rw----. 1 kbojens kbojens 0 29. Okt 18:35 test_simple_UnixFileLock_Path_0/a
-rw-rw----. 1 kbojens kbojens 0 29. Okt 18:35 test_simple_UnixFileLock_Path_current/a
-rw-rw----. 1 kbojens kbojens 0 29. Okt 18:35 test_simple_UnixFileLock_PureP0/a
-rw-rw----. 1 kbojens kbojens 0 29. Okt 18:35 test_simple_UnixFileLock_PurePcurrent/a
-rw-rw----. 1 kbojens kbojens 0 29. Okt 18:35 test_simple_UnixFileLock_str_0/a
-rw-rw----. 1 kbojens kbojens 0 29. Okt 18:35 test_simple_UnixFileLock_str_current/a
```

Do the tests run on your local machine? Is there anything different about Github's environment?

gaborbernat:
> Do the tests run on your local machine? Is there anything different about Github's environment?

They also fail on my local machine.

ghost:
> > Do the tests run on your local machine? Is there anything different about Github's environment?

> They also fail on my local machine.

As of now I have no idea why the test has this strange result with `0640` – all the tests run fine on my Fedora. I had a suspicion about the `umask` but even the most liberal setting changed nothing. I will investigate this further but this might take some time. If anyone has a clue or a hint I'd very much like to hear it.

ghost:
The problem was that the Github Runner has this `id`:
`uid=1001(runner) gid=121(docker) groups=121(docker),4(adm),101(systemd-journal)`

This lead to a problem with the umask. I have changed `_unix.py` and `_soft.py` that the umask is now changed temporarily for the operation. It now just fails for windows which is out of my expertise. I'll look into it …

ghost:
I have now modified the file permissions test to skip if the platform is `win32` as I have absolutely no clue about file permissions on this platform. This should be reviewed by someone with deep knowledge of the windows platform.

gaborbernat:
Don't skip the test please, we can't merge until we figure it out.

src/filelock/_soft.py
```diff
@@ -12,14 +12,17 @@ class SoftFileLock(BaseFileLock):
...
         # first check for exists and read-only mode as the open will mask this case as EEXIST
-        mode = (
+        flags = (
             os.O_WRONLY  # open for writing only
             | os.O_CREAT
             | os.O_EXCL  # together with above raise EEXIST if the file specified by filename exists
             | os.O_TRUNC  # truncate the file to zero byte
         )
+        mode = 0o660
+        umask = 0o002
```
gaborbernat, line 22: After looking into this seems chmod/umask is not supported on Windows. Calling those methods will not fail but Windows does not honour those settings as the underlying filesystem does not understand it. In light of that can we not set these on Windows? My proposal is to move the mode/umask shenanigans into a context manager here and not call it on Windows.

gaborbernat (changes requested):

src/filelock/_soft.py
ghost: I also did some research and tried several things on Windows via VirtualBox. File permissions are completely different and have nothing to do with those under Linux. I even asked on StackOverflow and got some hints:

https://stackoverflow.com/questions/69782278/os-open-windows-file-permissions-mode?noredirect=1#comment123348708_69782278

> My proposal is to move the mode/umask shenanigans into a context manager here and not call it on Windows.

I'm not sure that I understand your point. There already is a `_windows.py` with this:

```
if sys.platform == "win32":  # pragma: win32 cover
    import msvcrt

    class WindowsFileLock(BaseFileLock):
```

There are no changes to the mode or the (in Windows non-existing) umask. Please note that `mode` should also be renamed to `flags` in this file.

And this happens in `__init__.py`:

```
if sys.platform == "win32":  # pragma: win32 cover
    _FileLock: Type[BaseFileLock] = WindowsFileLock
```

Why would the setting of `umask+mode` have to be moved away? I'm still new to this project, so is there something I've overlooked?

gaborbernat: I was talking about the soft filelock implementation solely, not the platform specific locks.

ghost: I'll look into it, although it might take some time.

SATOSHI66 (approved):

SATOSHI66 (approved):

gaborbernat (changes requested):
Can you please check the failing tests?

ghost:
> Can you please check the failing tests?

Yes, I'll look into it this weekend. Sorry that it took so long.

ghost:
> Can you please check the failing tests?

I'm a little bit surprised. These changes should not have been approved. In https://github.com/tox-dev/py-filelock/pull/113#discussion_r739777614 we discussed the failing tests on Windows and you proposed to put all this into a context manager.

My current status is that I have successfully put all this into a context manager. The only remaining problem I have is with the exceptions, which now have to be raised within in the context manager.  I hope that I'll solve this problem until the next weekend.

ghost:
> Can you please check the failing tests?

Just a little update: I have successfully put the logic in a `@contextmanager`. But now some of the threaded tests are failing due to the generator being one time only. I am looking into the `@contextlib.asynccontextmanager` as a possible solution. As this is a rather deep rabbit hole I'll dive deeper into it between Christmas and New Year :)

gaborbernat:
> . But now some of the threaded tests are failing due to the generator being one time only.

What do you mean? 

> I am looking into the `@contextlib.asynccontextmanager` as a possible solution.

This is almost certainly not the path ahead.

ghost:
> > . But now some of the threaded tests are failing due to the generator being one time only.

> What do you mean?

For example the `test_threaded_shared_lock_obj` runs fine while `test_threaded_lock_different_lock_obj` fails for the SoftFileLock (UnixFileLock runs fine) with:

```
>           raise RuntimeError("generator didn't yield") from None
E           RuntimeError: generator didn't yield
```

There are some other tests still failing and I'm looking into it.

gaborbernat:
@kbojens do you still intend to finish this one?

ghost:
> @kbojens do you still intend to finish this one?

Yes, I'll take a final look at the problem this weekend and will then decide how to proceed.

ghost:
> Yes, I'll take a final look at the problem this weekend and will then decide how to proceed.

Ok, I'm now pretty sure that I now have a solution for all the problems and will keep working on this.

PR #119, opened 2021-11-15
Fix misspelled keyword argument `poll_interval` for method `acquire`

### LINKED ISSUE #62: poll_intervall is a misspelling
poll_intervall should be spelled poll_interval (but leave the old one for compatibility!)

XuehaiPan:
Fix misspelled keyword argument `poll_interval` for method `acquire`. Fixes #62.

I tried my best to add the backward compatibility layer in this PR to not break existing codebases.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -111,7 +111,7 @@ class BaseFileLock(ABC):
         self,
         timeout: Optional[float] = None,
         poll_interval: float = 0.05,
-        poll_intervall: Optional[float] = None,  # kept for backward compatibility, will be removed in a future release
+        poll_intervall: Optional[float] = None,
     ) -> AcquireReturnProxy:
         """
         Try to acquire the file lock.
@@ -119,6 +119,7 @@ class BaseFileLock(ABC):
         :param timeout: maximum wait time for acquiring the lock, ``None`` means use the default :attr:`~timeout` is and
          if ``timeout < 0``, there is no timeout and this method will block until the lock could be acquired
         :param poll_interval: interval of trying to acquire the lock file
+        :param poll_intervall: deprecated, kept for backwards compatibility, use ``poll_interval`` instead
         :raises Timeout: if fails to acquire lock within the timeout period
         :return: a context object that will unlock the file when the context is exited
 
@@ -137,23 +138,17 @@ class BaseFileLock(ABC):
 
         .. versionchanged:: 2.0.0
 
-            This method returns now a *proxy* object instead of *self*,
-            so that it can be used in a with statement without side effects.
-
+            This method returns now a *proxy* object instead of *self*, so that it can be used in a with statement \
+            without side effects.
 
         """
         # Use the default timeout, if no timeout is provided.
         if timeout is None:
             timeout = self.timeout
 
-        # Remove this and argument `poll_intervall` above in a future release
         if poll_intervall is not None:
-            warnings.warn(
-                "poll_intervall is a deprecated misspelled keyword "
-                "and will be removed in a future release. "
-                "Please use poll_interval instead.",
-                DeprecationWarning,
-            )
+            msg = "use poll_interval instead of poll_intervall"
+            warnings.warn(msg, DeprecationWarning)
             poll_interval = poll_intervall
 
         # Increment the number right at the beginning. We can still undo it, if something fails.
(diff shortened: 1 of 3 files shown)


src/filelock/_api.py
```diff
@@ -106,13 +106,13 @@ def is_locked(self) -> bool:
         """
         return self._lock_file_fd is not None
 
-    def acquire(self, timeout: Optional[float] = None, poll_intervall: float = 0.05) -> AcquireReturnProxy:
+    def acquire(self, timeout: Optional[float] = None, poll_interval: float = 0.05, **kwargs) -> AcquireReturnProxy:
```
gaborbernat: Please don't accept everything, the only thing we do need to accept is the old variant.

src/filelock/_api.py
```diff
@@ -140,6 +140,17 @@ def acquire(self, timeout: Optional[float] = None, poll_intervall: float = 0.05)
         if timeout is None:
             timeout = self.timeout
 
+        # TODO: remove this and `**kwargs` above in a future release
```
gaborbernat: No TODO please

gaborbernat (changes requested):

src/filelock/_api.py
XuehaiPan: At lines 151-152, it will raise `TypeError` when the user provides something other than the old variant.

gaborbernat: Yeah, I saw. But we don't need that, just express the intent via the interface, not fixup code.

src/filelock/_api.py
```diff
@@ -140,6 +140,17 @@ def acquire(self, timeout: Optional[float] = None, poll_intervall: float = 0.05)
         if timeout is None:
             timeout = self.timeout
 
+        # Remove this and `**kwargs` above in a future release
+        if "poll_intervall" in kwargs:
+            _LOGGER.warning(
```
gaborbernat: This should be done via the warnings module

src/filelock/_api.py
```diff
@@ -106,13 +107,18 @@ def is_locked(self) -> bool:
         """
         return self._lock_file_fd is not None
 
-    def acquire(self, timeout: Optional[float] = None, poll_intervall: float = 0.05) -> AcquireReturnProxy:
+    def acquire(
+        self,
+        timeout: Optional[float] = None,
+        poll_interval: float = 0.05,
```
gaborbernat, line 113: Why are you changing the interface type here?

gaborbernat (changes requested):
Can you please add a test to validate this?

XuehaiPan:
I think we can reimplement this by using decorators without changing the signature of the method.

gaborbernat:
> I think we can reimplement this by using decorators without changing the signature of the method.

I think the current implementation is fine, no need for decorators.

gaborbernat (approved):

gaborbernat:
Thanks for your contribution.

ofek:
I don't know where it's coming from (maybe virtualenv?) but I'm getting a lot of warnings:

```
tests/cli/build/test_build.py: 10 warnings
tests/cli/env/test_create.py: 45 warnings
tests/cli/env/test_prune.py: 6 warnings
tests/cli/env/test_remove.py: 12 warnings
tests/cli/publish/test_publish.py: 42 warnings
tests/cli/run/test_run.py: 18 warnings
tests/dep/test_core.py: 24 warnings
tests/venv/test_core.py: 21 warnings
  /home/runner/.local/share/hatch/env/virtual/hatch-kmtIAcBK/hatch/lib/python3.9/site-packages/filelock/_api.py:151: DeprecationWarning: use poll_interval instead of poll_intervall
    warnings.warn(msg, DeprecationWarning)
```

XuehaiPan:
> I don't know where it's coming from (maybe virtualenv?) but I'm getting a lot of warnings:

@ofek Can you try to run `python -W error xxx.py` or `pytest ... -W error` to see what module uses the old argument `poll_intervall`?

ofek:
Thanks, found it!

https://github.com/pypa/virtualenv/blob/623189a00aa96c805990fd12dc883c69e251f4b2/src/virtualenv/util/lock.py#L31

ofek:
https://github.com/pypa/virtualenv/pull/2237

PR #121, opened 2021-11-20
Add `stacklevel` to deprecation warnings for argument name change

XuehaiPan:
Changes:

1. Add `*` to function signature to mark the deprecated argument `poll_intervall` as keyword only.
2. Add `stacklevel=2` for deprecation warnings, show the file and lineno for the user's code rather than this module (filelock).

    From https://github.com/tox-dev/py-filelock/pull/119#issuecomment-974607426:

    > I don't know where it's coming from (maybe virtualenv?) but I'm getting a lot of warnings

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -111,6 +111,7 @@ class BaseFileLock(ABC):
         self,
         timeout: Optional[float] = None,
         poll_interval: float = 0.05,
+        *,
         poll_intervall: Optional[float] = None,
     ) -> AcquireReturnProxy:
         """
@@ -138,8 +139,8 @@ class BaseFileLock(ABC):
 
         .. versionchanged:: 2.0.0
 
-            This method returns now a *proxy* object instead of *self*, so that it can be used in a with statement \
-            without side effects.
+            This method returns now a *proxy* object instead of *self*,
+            so that it can be used in a with statement without side effects.
 
         """
         # Use the default timeout, if no timeout is provided.
@@ -148,7 +149,7 @@ class BaseFileLock(ABC):
 
         if poll_intervall is not None:
             msg = "use poll_interval instead of poll_intervall"
-            warnings.warn(msg, DeprecationWarning)
+            warnings.warn(msg, DeprecationWarning, stacklevel=2)
             poll_interval = poll_intervall
 
         # Increment the number right at the beginning. We can still undo it, if something fails.
--- a/tests/test_filelock.py
+++ b/tests/test_filelock.py
@@ -2,6 +2,7 @@ import logging
 import sys
 import threading
 from contextlib import contextmanager
+from inspect import getframeinfo, stack
 from pathlib import Path, PurePath
 from stat import S_IWGRP, S_IWOTH, S_IWUSR
 from types import TracebackType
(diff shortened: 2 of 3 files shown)


src/filelock/_api.py
```diff
@@ -111,6 +111,7 @@ def acquire(
         self,
         timeout: Optional[float] = None,
         poll_interval: float = 0.05,
+        *,
```
gaborbernat, line 114: Why this unrelated change?

gaborbernat (changes requested):

src/filelock/_api.py
gaborbernat: > Add * to function signature to mark the deprecated argument poll_intervall as keyword only.

I don't think we should do this, that would be a breaking change, please remove that. And add a test for the stack level.

XuehaiPan: > I don't think we should do this, that would be a breaking change.

It would not be a breaking change because the positional argument will be captured by `poll_interval`. In our case, the user will receive a deprecation warning when they use `acquire` with the old keyword `lock.acquire(poll_intervall=0.05)`.

Without marking `poll_intervall` as keyword-only, the user can write:

```python
lock.acquire(None, 0.05, 0.05)
```

XuehaiPan: > And add a test for the stack level.

`pytest` does not support checking `stacklevel` yet (see pytest-dev/pytest#4343). I'll find a workaround.

tests/test_filelock.py
```diff
@@ -357,5 +358,11 @@ def test_poll_intervall_deprecated(lock_type: Type[BaseFileLock], tmp_path: Path
...
-    with pytest.deprecated_call(match="use poll_interval instead of poll_intervall"):
-        lock.acquire(poll_intervall=0.05)
+    with pytest.deprecated_call(match="use poll_interval instead of poll_intervall") as checker:
+        lock.acquire(poll_intervall=0.05)  # call deprecated keyword argument
+        frameinfo = getframeinfo(stack()[0][0])  # get frameinfo of current file and lineno (+1 than the above lineno)
+        for warning in checker:
+            if warning.filename == frameinfo.filename and warning.lineno + 1 == frameinfo.lineno:
+                break
+        else:
+            pytest.fail("No warnings of stacklevel=2 matching.")
```
gaborbernat: Can you please mark this as pragma no cover?

gaborbernat (changes requested):

tests/test_filelock.py
```diff
@@ -357,5 +358,11 @@ def test_poll_intervall_deprecated(lock_type: Type[BaseFileLock], tmp_path: Path
     lock_path = tmp_path / "a"
     lock = lock_type(str(lock_path))
 
-    with pytest.deprecated_call(match="use poll_interval instead of poll_intervall"):
-        lock.acquire(poll_intervall=0.05)
+    with pytest.deprecated_call(match="use poll_interval instead of poll_intervall") as checker:
+        lock.acquire(poll_intervall=0.05)  # call deprecated keyword argument
+        frameinfo = getframeinfo(stack()[0][0])  # get frameinfo of current file and lineno (+1 than the above lineno)
+        for warning in checker:
```
gaborbernat, line 364: Can you make sure the diff cover is 100%, please?

gaborbernat (changes requested):

gaborbernat (approved):

PR #128, opened 2022-01-03
Enable use as a context decorator

ghost:


docs/index.rst
```diff
@@ -67,6 +67,14 @@ locks:
     finally:
         lock.release()
 
+
+    @lock
+    def decorated():
+        open(file_path, "a").write("You're a decorated Jedi!")
```
gaborbernat: just print the message and avoid the file open

gaborbernat (changes requested):

docs/index.rst
ghost: Fixed

gaborbernat (approved):

FFY00 (approved):

PR #135, opened 2022-02-14
Raise NotImplementedError if using wrong class

### LINKED ISSUE #102: 3.3.0 breaks pylint
Since the last update, I am getting the following warnings. I suspect this has to do with the fact that pylint cannot deduce that the correct instance will not be superclassed by ABC (if-else definition of classes).

```python
import filelock

with filelock.FileLock("abc.lock"):
    print("hi")

```

```bash
pylint bla.py 
************* Module bla
bla.py:1:0: C0114: Missing module docstring (missing-module-docstring)
bla.py:3:5: E0110: Abstract class 'WindowsFileLock' with abstract methods instantiated (abstract-class-instantiated)
bla.py:3:5: E0110: Abstract class 'UnixFileLock' with abstract methods instantiated (abstract-class-instantiated)

```

vonschultz:
Instead of leaving abstract methods without definition when running on the other platform (Unix vs Windows), define them and raise a `NotImplementedError` if called on the wrong platform.

This fixes pylint warnings such as
```
   Abstract class 'WindowsFileLock' with abstract methods instantiated
   Abstract class 'UnixFileLock' with abstract methods instantiated
```

Fixes tox-dev/py-filelock#102

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,8 +1,8 @@
 Changelog
 =========
 
-Future
-------
+v3.6.0 (2022-02-17)
+-------------------
 - Fix pylint warning "Abstract class :class:`WindowsFileLock <filelock.WindowsFileLock>` with abstract methods instantiated"
   :pr:`135` - by :user:`vonschultz`
 - Fix pylint warning "Abstract class :class:`UnixFileLock <filelock.UnixFileLock>` with abstract methods instantiated"
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+Future
+------
+- Fix pylint warning "Abstract class :class:`WindowsFileLock <filelock.WindowsFileLock>` with abstract methods instantiated"
+  :pr:`135` - by :user:`vonschultz`
+- Fix pylint warning "Abstract class :class:`UnixFileLock <filelock.UnixFileLock>` with abstract methods instantiated"
+  :pr:`135` - by :user:`vonschultz`
+
 v3.4.2 (2021-12-16)
 -------------------
 - Drop support for python ``3.6``
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -14,6 +14,10 @@ from ._error import Timeout
 _LOGGER = logging.getLogger("filelock")
 
 
+class PlatformMismatchError(Exception):
+    """Exception raised when attempting Unix locks on Windows, or vice versa."""
+
+
 # This is a helper class which is returned by :meth:`BaseFileLock.acquire` and wraps the lock to make sure __enter__
 # is not called twice when entering the with statement. If we would simply return *self*, the lock would be acquired
 # again in the *__enter__* method of the BaseFileLock, but not released again automatically. issue #37 (memory leak)
@@ -236,4 +240,5 @@ class BaseFileLock(ABC):
 __all__ = [
     "BaseFileLock",
     "AcquireReturnProxy",
+    "PlatformMismatchError",
 ]
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -14,10 +14,6 @@ from ._error import Timeout
 _LOGGER = logging.getLogger("filelock")
 
 
-class PlatformMismatchError(Exception):
-    """Exception raised when attempting Unix locks on Windows, or vice versa."""
-
-
 # This is a helper class which is returned by :meth:`BaseFileLock.acquire` and wraps the lock to make sure __enter__
 # is not called twice when entering the with statement. If we would simply return *self*, the lock would be acquired
 # again in the *__enter__* method of the BaseFileLock, but not released again automatically. issue #37 (memory leak)
@@ -240,5 +236,4 @@ class BaseFileLock(ABC):
 __all__ = [
     "BaseFileLock",
     "AcquireReturnProxy",
-    "PlatformMismatchError",
 ]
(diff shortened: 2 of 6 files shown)


gaborbernat (changes requested):
I prefer in this case the class to remain abstract. This is a clearly faulty warning in pylint and the solution should be to change pylint check than modify our code. #102 was closed for a reason: the issue is not with us, is with pylint.

vonschultz:
I don't see why you'd want the class to remain abstract, regardless of what pylint may or may not think of it. An abstract class is a class for which the client can be expected to provide an implementation, which doesn't make sense in this case — the client _isn't_ supposed to implement a WindowsFileLock if he's running on Unix, so why would you want it to be an abstract class?

I took your comment on #102 as an indication that you _were_ willing to consider a solution "that pleases both the static checker + type checker + documentation generation." Wouldn't this work?

gaborbernat:
> I don't see why you'd want the class to remain abstract,

I mean because it's abstract and lacking methods that should be implemented? 😂

I guess our definition of what abstract class differs significantly. In my book an abstract class is a class that shouldn't be created as it does not have some of its required methods defined (for whatever reason). The current design satisfies that requirement.

gaborbernat:
> took your comment on #102 as an indication that you were willing to consider a solution \"that pleases both the static checker + type checker + documentation generation.\

I did consider it, but I'm not in love at the moment with the proposal 🤔

vonschultz:
Ah, OK, fair enough. I did consider raising an issue with Pylint instead, but, well... with my interpretation of what an abstract class is, Pylint's behavior seems rather reasonable to me. If an abstract class defines an interface for client code to implement, if that is what we mean by "abstract class," then it should be fair to assume that being abstract or not isn't a runtime property: If there's any branch where the class is abstract, then it's defining an interface, and then you shouldn't use it without subclassing it first.

Let me know if there's any way to tweak my proposal to make it more to your liking.

src/filelock/_api.py
```diff
@@ -14,6 +14,10 @@
 _LOGGER = logging.getLogger("filelock")
 
 
+class PlatformMismatchError(Exception):
```
gaborbernat: Let's remove this class 👍

src/filelock/_unix.py
```diff
@@ -2,18 +2,23 @@
...
 #: a flag to indicate if the fcntl API is available
 has_fcntl = False
 if sys.platform == "win32":  # pragma: win32 cover
 
-    class UnixFileLock(BaseFileLock, ABC):
+    class UnixFileLock(BaseFileLock):
         """Uses the :func:`fcntl.flock` to hard lock the lock file on unix systems."""
 
+        def _acquire(self) -> None:
+            raise PlatformMismatchError()
```
gaborbernat: Let's just change this and the others

```raise NotImplementedError```

**Review by gaborbernat — CHANGES_REQUESTED** [review]
You'll want to add tests too 👍  and a changelog entry.

File: docs/changelog.rst [hunk 3]
```diff
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+Future
```
gaborbernat: Mark it the next minor and today's date 👍

tests/test_filelock.py
```diff
@@ -368,3 +369,35 @@ def test_poll_intervall_deprecated(lock_type: type[BaseFileLock], tmp_path: Path
...
+    assert not inspect.isabstract(
+        WindowsFileLock
+    ), "WindowsFileLock must not be an abstract class, or pylint will complain in client code"
+    assert inspect.isabstract(BaseFileLock)
+
+    with pytest.raises(NotImplementedError):
+        lock.acquire()
+
+    with pytest.raises(NotImplementedError):
+        lock._release()
```
gaborbernat: Can you make this instead one parametrized test (based on platform)

gaborbernat (changes requested):

tests/test_filelock.py
vonschultz: Do you mean with `@pytest.mark.parametrize()`? I don't really see how that would be done here. I can make it a single test fairly easily, though.

tests/test_filelock.py
```diff
@@ -385,26 +385,13 @@ def decorated_method() -> None:
...
 
 
-@pytest.mark.skipif(sys.platform != "win32", reason="Tests behavior of UnixFileLock on Windows systems")
-def test_unix_lock_on_windows(tmp_path: Path) -> None:
-    lock = UnixFileLock(str(tmp_path / "lockfile"))
+def test_wrong_platform(tmp_path: Path) -> None:
+    lock_type = UnixFileLock if sys.platform == "win32" else WindowsFileLock
+    lock = lock_type(str(tmp_path / "lockfile"))
 
     assert not inspect.isabstract(  # noqa: SC200
```
gaborbernat: a better test here would be to instantiate the class and test that calling the acquire/release raises NotImplementedError 🤔

gaborbernat (changes requested):

tests/test_filelock.py
vonschultz: I do instantiate the class and test that acquire/release raises NotImplementedError.

vonschultz: We could skip the isabstract() part entirely, since it's implicit in that we instantiate the class. Though saying it explicitly makes it clearer that the non-abstractness is a desired outcome in itself, and is being tested on purpose.

gaborbernat: sure you can do that too, but then still need to instantiate to check that calling those methods raises; up to you; and yes I'd make it parametrized personally, but you can get by with if statements too

gaborbernat (approved):

PR #136, opened 2022-02-16
Use time.monotonic() for calculating timeouts.

avian2:
time.time() is affected by system clock updates (for example, by the NTP
daemon). This can cause timeouts to be unpredictable if clock jumps forwards or
backwards while code is waiting to acquire a lock.

time.monotonic() returns a time value that is not affected by clock updates and
does not have this issue. In Python 3.5 and later it is available on all systems[1].

[1] https://docs.python.org/3/library/time.html#time.monotonic

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,7 +3,7 @@ Changelog
 
 v3.5.1 (2022-02-16)
 -------------------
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
Can you please add a changelog entry too?

gaborbernat (approved):

PR #139, opened 2022-03-23
Workaround for a bug in the Google Drive app

dirkgr:
The Google Drive app has some interesting behavior if it tries to sync the directory that contains the lock file. If the file is empty, it will replace the file with another empty file of equal name. The lock stays with the original inode. Another process can lock the same file again, because it now gets a different inode.

This does not happen with files that aren't empty, so I changed the code to write a small explanation into the lock file when it first gets created.

### DIFF
--- a/src/filelock/_unix.py
+++ b/src/filelock/_unix.py
@@ -32,16 +32,27 @@ else:  # pragma: win32 no cover
 
         def _acquire(self) -> None:
             open_mode = os.O_RDWR | os.O_CREAT
-            fd = os.open(self._lock_file, open_mode)
-            try:
-                if os.fstat(fd).st_size == 0:
-                    os.write(fd, b"Lock files must not be empty, or the Google Drive app will replace them.")
-                    os.fsync(fd)
-                fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-            except OSError:
-                os.close(fd)
-            else:
-                self._lock_file_fd = fd
+            while True:
+                # Make sure we have a non-empty file.
+                fd = os.open(self._lock_file, open_mode)
+                try:
+                    if os.fstat(fd).st_size == 0:
+                        os.write(fd, b"Lock files must not be empty, or the Google Drive app will replace them.")
+                finally:
+                    os.close(fd)
+
+                fd = os.open(self._lock_file, open_mode)
+                try:
+                    if os.fstat(fd).st_size == 0:
+                        # Looks like Google Drive already replaced the file. Try again.
+                        os.close(fd)
+                        continue
+                    fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                except OSError:
+                    os.close(fd)
+                else:
+                    self._lock_file_fd = fd
+                break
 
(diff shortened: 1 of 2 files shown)


src/filelock/_unix.py
```diff
@@ -31,9 +31,12 @@ class UnixFileLock(BaseFileLock):
         """Uses the :func:`fcntl.flock` to hard lock the lock file on unix systems."""
 
         def _acquire(self) -> None:
-            open_mode = os.O_RDWR | os.O_CREAT | os.O_TRUNC
+            open_mode = os.O_RDWR | os.O_CREAT
             fd = os.open(self._lock_file, open_mode)
             try:
+                if os.fstat(fd).st_size == 0:
```
gaborbernat: I don't think we can do this change, you've introduced a race condition here now where if Google Drive syncs between open and fsync the error still manifest; and removing the truncate makes the lock file not unique. IMHO the problem should be fixed on Google Drive side, not here.

dirkgr: I fixed the race condition.

> removing the truncate makes the lock file not unique

I don't understand that. What do you mean?

Of course it should be fixed by Google, but you know as well as I do that that'll never happen.

gaborbernat: Did you file a bug report for it? I'm inclined to leave it as is and not add such an ugly hack to make GDrive work.

dirkgr: I did not file a bug for it. There is no publicly accessible bug tracker, and regular Customer Support is going to be a waste of time. I pinged them on Twitter. For some companies that works, but not usually Google.

Can we at least get rid of truncation, so I can roll my own workaround in my code? As it is, this is unfixable when you're using filelock.

gaborbernat: I'll let other maintainers chime in, but from my side I'll give this a -1 👍

dirkgr: It's your codebase, and I won't pretend this is a pretty fix. But I want to be clear here: Without this fix, `filelock` will grant an exclusive lock to multiple processes at the same time. The lock will be ignored, silently, with whatever consequences that brings to the application's data. This happens if you have a (from the user's perspective) completely unrelated app installed on your machine.

Google Drive has over 1B users. Not all of them run on Macs, and not all of them run the app, but the top number is so large it's practically guaranteed that this happens in the wild.

gaborbernat: Just to be clear, the bug still is not in our code base but with Google Drive. If it impacts millions of users surely they can fix their own bugs.

dirkgr: They absolutely should. But they won't. Quote from a friend of mine that works at Google (about a different matter): "Unless it affects 1M people or an entire country, don't bother me with it.". 

From a user's perspective, the bug will show up in apps that use `filelock`. I happened to stumble on this connection because of the way we use it. Most people won't trace it back.

gaborbernat: Guess we have to leave it there until a million people run into it ☺️

gaborbernat:
Thanks for your contribution but seems this is a bug for Google Drive to fix, so for no we'll not adopt it.

Issue #57, opened 2019-12-12
Timeout-free lock acquisition


chrysn:
For an application that has no hopes of waiting for a lock to become usable (in my case, avoiding the simultaneous execution of two server processes on the same state file), I'd need locks that fail immediately when held by another process.

The trivial timeout value I'd like to pass, `0`, has a different meaning, though: "poll indefinitely"; the `None` value that would lend itself to the desired behavior is also used differently (meaning "use the default for that lock").

My current workaround is to just pass `0.001`, and it works. (Given the user story there is "start the server, see that it fails immediately", that's immediate enough).

It would be nice to have a documented way of getting immediate-return behavior. For API stability reasons, neither redefining `0` or `None` are viable. One way to introduce the desired behavior is to have a `filelock.RETURN_IMMEDIATELY` value (an `object()`, a string or even `0.000000001`) that can be passed in as a timeout argument to indicate that return or error should show immediately.

gaborbernat:
Hello, if you make a PR for this (with tests) we would be happy to review, thanks!

axoroll7:
Is your issue resolved now ? If it is the case, can you close the issue please.

chrysn:
I tried to reproduce the issue with the latest release, and it appears to be well-behaved w/rt a timeout=0. Then again, so does the release that was current at the time I've opened the issue. For lack of a way to reproduce it, and as it is apparently not relevant any more, I'm closing it.

Thanks for your input, and for providing this library.

PR #142, opened 2022-05-03
Add ability to return immediately when a lock cannot be obtained inst…

myheroyuki:
Addresses issue #57

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -9,6 +9,7 @@ from abc import ABC, abstractmethod
 from threading import Lock
 from types import TracebackType
 from typing import Any
+from xmlrpc.client import Boolean
 
 from ._error import Timeout
 
@@ -116,7 +117,7 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         poll_interval: float = 0.05,
         *,
         poll_intervall: float | None = None,
-        blocking=True,
+        blocking: bool = True,
     ) -> AcquireReturnProxy:
         """
         Try to acquire the file lock.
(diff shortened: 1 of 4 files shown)


jugmac00:
There is a typo in `ImmediateAquireError`  - I think this should be `ImmediateAcquireError`

FFY00 (commented):
Wouldn't `blocking`/`nonblocking` be a more standard name?

Eg. https://docs.python.org/3/library/threading.html#threading.Lock.acquire

myheroyuki:
I chose return_immediately because it was suggested in the original issue, but I agree blocking/nonblocking is the more standard convention. I updated the PR to change to blocking.

jugmac00:
There is one flake 8 issue.

src/filelock/_api.py
```diff
@@ -116,6 +116,7 @@ def acquire(
         poll_interval: float = 0.05,
         *,
         poll_intervall: float | None = None,
+        blocking=True,
```
gaborbernat: A better solution wouldn't just use the value 0 to mean non-blocking? Do we need this extra argument?

gaborbernat (changes requested):

src/filelock/_api.py
myheroyuki: As discussed in the issue #57, a timeout value of 0 currently means poll-indefinitely. Adding an additional argument makes it's functionality more clear without breaking backwards compatibility.

gaborbernat:
You'll need to fix the CI 👍

gaborbernat (approved):

PR #164, opened 2022-08-29
Use license_files instead of license_file

felixonmars:
Fix the following warning:

```
/usr/lib/python3.10/site-packages/setuptools/config/setupcfg.py:459: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.
  warnings.warn(msg, warning_class)
```

Ref: https://github.com/pypa/setuptools/pull/2620


felixonmars:
pre-commit-ci is breaking the fix here :(

PR #177, opened 2022-12-05
Fix mypy fails understanding FileLock

gaborbernat:
Signed-off-by: Bernát Gábor <bgabor8@bloomberg.net>


nanonyme:
I'm curious, why are you setting so restricted versions for build dependencies? Was there some bug you were trying to avoid?

gaborbernat:
> I'm curious, why are you setting so restricted versions for build dependencies? Was there some bug you were trying to avoid?

Because we're only testing the latest versions in the CI, anything else is untested and unsupported. That's how open source works.

Issue #112, opened 2021-10-25
File permission of lock file


ghost:
The creation of a lockfile with …

`lock        = FileLock("/var/lock/foo.lock")`

… leads to these file permissions: `-rwxr-xr-x` 

Is there any way to prevent that the lock file becomes an executable with root ownership?

(Version: 3.0.12-2 in Ubuntu 20.04)

gaborbernat:
Can't you just change the file permissions after the lock file is created?

ghost:
Sure. But this is a potential security problem and should not happen at all. Creating files which are world wide executable is quite the opposite to the principle of least surprise.

gaborbernat:
Feel free to open a PR with your proposed solution.

ghost:
Ok. I just have a question about the code:

```
mode = (
            os.O_WRONLY  # open for writing only
            | os.O_CREAT
            | os.O_EXCL  # together with above raise EEXIST if the file specified by filename exists
            | os.O_TRUNC  # truncate the file to zero byte
        )
        try:
            fd = os.open(self._lock_file, mode)
```

If I understand the documentation of `os.open` and the C runtime docs correctly these parameters should be called `flags` instead of `mode` and `mode` would then instead define the file permissions like `S_IRUSR`, or am I missing something?

JustAnotherArchivist:
Your reading of the documentation is correct. The value of `mode` in the filelock code is passed into `os.open`'s `flags` argument. This doesn't matter because `os.open` never sees the name of the variable being passed in, but I agree that `mode` should be renamed to `flags` to be less confusing. And then the third argument (= `mode` keyword argument) to `os.open` could be passed in whichever notation is preferred (`os.S_IRUSR | os.S_IWUSR` or `0o600`, for example). I'm not sure what platform-specific tweaks might be necessary, especially on the Windows implementation.

axoroll7:
> Sure. But this is a potential security problem and should not happen at all. Creating files which are world wide executable is quite the opposite to the principle of least surprise.

The lock file is just an empty file ? Why is it a security problem ? I understand your motivation, this file should not be executable. It would be nice if you have an explication, because you seem to know the potential security problems.

jahrules:
@gaborbernat - I believe the recent PRs should resolve this one. Please let me know if you disagree or if you believe any additional work is needed here.

PR #192, opened 2023-02-02
added multiuser support and associated tests

jahrules:
new option FileLock(multiuser = True) (defaults to False):

results in Lock File being created with rw-rw-rw- permissions so that all users on the system can use the lockfile.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,9 @@
 Changelog
 =========
+v3.10.0 (2023-03-15)
+-------------------
+- Add support for lockfile file mode :pr:`192 - by :user:`jahrules`.
+
 v3.9.1 (2023-03-14)
 -------------------
 - Use ``time.perf_counter`` instead of ``time.monotonic`` for calculating timeouts.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,7 @@ Changelog
 =========
 v3.10.0 (2023-03-15)
 -------------------
-- Add support for lockfile file mode :pr:`192 - by :user:`jahrules`.
+- Add support for explicit file modes for lockfiles :pr:`192 - by :user:`jahrules`.
 
 v3.9.1 (2023-03-14)
 -------------------
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -43,7 +43,7 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         self,
         lock_file: str | os.PathLike[Any],
         timeout: float = -1,
-        multiuser: bool = False,
+        multi_user: bool = False,
     ) -> None:
         """
         Create a new lock object.
@@ -64,8 +64,8 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         # The default timeout value.
         self._timeout: float = timeout
 
-        # The multiuser flag
-        self._multiuser: bool = multiuser
+        # The multi_user flag
+        self._multi_user: bool = multi_user
 
         # We use this lock primarily for the lock counter.
         self._thread_lock: Lock = Lock()
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -52,7 +52,7 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         :param timeout: default timeout when acquiring the lock, in seconds. It will be used as fallback value in
         the acquire method, if no timeout value (``None``) is given. If you want to disable the timeout, set it
         to a negative value. A timeout of 0 means, that there is exactly one attempt to acquire the file lock.
-        : param mode: file permissions for the lockfile to have. default: 0o644 (-rw-r--r--)
+        : param mode: file permissions for the lockfile.
         """
         # The path to the lock file.
         self._lock_file: str = os.fspath(lock_file)
@@ -182,7 +182,6 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
                         previous_umask = os.umask(0)
                         self._acquire()
                         os.umask(previous_umask)  # reset umask to initial value
-
                 if self.is_locked:
                     _LOGGER.debug("Lock %s acquired on %s", lock_id, lock_filename)
                     break
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -52,7 +52,7 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         :param timeout: default timeout when acquiring the lock, in seconds. It will be used as fallback value in
         the acquire method, if no timeout value (``None``) is given. If you want to disable the timeout, set it
         to a negative value. A timeout of 0 means, that there is exactly one attempt to acquire the file lock.
-        : param multiuser: if True, make the lock file permissions rw/rw/rw for all users
+        : param mode: file permissions for the lockfile to have. default: 0o644 (-rw-r--r--)
         """
         # The path to the lock file.
         self._lock_file: str = os.fspath(lock_file)
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -39,7 +39,12 @@ class AcquireReturnProxy:
 class BaseFileLock(ABC, contextlib.ContextDecorator):
     """Abstract base class for a file lock object."""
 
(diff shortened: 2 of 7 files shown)


jahrules:
flake8 doesnt like the import filemode in test_filelock.py that I added to check the filemode with this addition. Aside from that; all tests should pass now.

Apologies for the gap from the original PR to these fixes. It was some time before I could access a windows box to debug.

jahrules:
I wanted to post perhaps a few clarifications.

This PR addresses requests noted in the following issues:
https://github.com/tox-dev/py-filelock/issues/112
https://github.com/tox-dev/py-filelock/issues/70

As mentioned in issue #112 I updated the variable name from mode to flags - which is a finding that I made independently before finding it mentioned in the aforementioned issue.

gaborbernat (changes requested):
Add changelog entry here https://github.com/tox-dev/py-filelock/blob/main/docs/changelog.rst

src/filelock/_api.py
```diff
@@ -39,14 +39,20 @@ def __exit__(
 class BaseFileLock(ABC, contextlib.ContextDecorator):
     """Abstract base class for a file lock object."""
 
-    def __init__(self, lock_file: str | os.PathLike[Any], timeout: float = -1) -> None:
+    def __init__(
+        self,
+        lock_file: str | os.PathLike[Any],
+        timeout: float = -1,
+        multi_user: bool = False,
```
gaborbernat: I'm not sure about this interface. Should we just expose access_control instead and allow them to set it to whatever they want?

gaborbernat (changes requested):

src/filelock/_api.py
jahrules: I'm open to this idea. My only concern is that when setting permissions for "group" or "all" you also have to set os.umask(0) so a change to that interface will also require extra checking for the requested permissions to set the umask if necessary

gaborbernat: I'm alright to make that check 👍

jahrules:
I have some additional information on this.

The default to mode in os.open is 511 (0o777 in octal). So the r/w bits are
set by default. The only reason that the default mode doesn't make the
files globally r/w is the lack of the os.umask call. So, even with exposing
the mode directly, I believe we need a way for users to specify if they
intend the lock to be multiuser or not.

Maybe a solution would to both allow the user to specify their own mode and
specify if multi-user is intended but this may lead to more confusion.

Open to thoughts on this.

jahrules:
Another potential solution would be to make the filelock default mode 0o700
but I'm not comfortable making the call on if we should deviate from the os
open default or not.

gaborbernat:
> I believe we need a way for users to specify if they intend the lock to be multiuser or not.

But there's no such thing as multi user. Is just mangles to acces to the lock created. It just happens that a given combination of flags enable multiple users to use the same lock file.

jahrules:
Well yes but it's the combination of flags plus umask. Typical default for
umask (at least on unix based OSes) is 0o133 to deny write to group/all.

So even though the os.open has a default mode of 0o777; the umask default
denies write making it effectively 0o644 which is why "other users" can't
use the same lock.

So to distill this to the core; I want to ask filelock users "do you want
other users to use this lock?" and modify umask to permit this.

Modifying mode and umask is pretty advanced for most users which is why
Python put that implementation into os.open and didn't expose it in the
standard "open".

Maybe the best option is to leave mode alone as I think it highly unlikely
that users would need granular permissions.

The three options I can reasonably see people wanting are "self" locks (as
they are today), "group" locks (where the "group" bit is set to write
(umask 0o113), and "global" locks (umask 0o111). The use case for the 3rd
would be like if an application user and human user both needed to use the
lock. The difference between 2 and 3 is, however, minimal.

src/filelock/_unix.py
```diff
@@ -31,8 +31,14 @@ class UnixFileLock(BaseFileLock):
         """Uses the :func:`fcntl.flock` to hard lock the lock file on unix systems."""
 
         def _acquire(self) -> None:
-            open_mode = os.O_RDWR | os.O_CREAT | os.O_TRUNC
-            fd = os.open(self._lock_file, open_mode)
+            open_flags = os.O_RDWR | os.O_CREAT | os.O_TRUNC
+            if self._multi_user is True:
+                os.umask(0)
```
gaborbernat: Shouldn't this be in a try/finally, so after creation we restore the previous value? Also do we need this conditionality? Can't we just always set umask to 0 before creation and restore after? Technically if the umask to set to something non zero, the open mode 511 will not be respected. Also this umask thing is repeated across files so any reason can't be a contextmanager in base?

gaborbernat (changes requested):

src/filelock/_unix.py
jahrules: agree completely. I would suggest that by going that route, the default should be 0o644 so that the 'net effect' of this change is that the default behavior remains the same. Let me know if you would prefer something else.

Also, I think that I will set the umask to 0o111 instead of 0; since there is no reason that the lock files should be executable.

Changes in progress.

src/filelock/_api.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         self,
         lock_file: str | os.PathLike[Any],
         timeout: float = -1,
-        multi_user: bool = False,
+        mode: int = 0o644,
```
gaborbernat, line 46: Add mode docstring param please 👍

src/filelock/_api.py
```diff
@@ -198,6 +199,9 @@ def acquire(
             with self._thread_lock:
                 self._lock_counter = max(0, self._lock_counter - 1)
             raise
+        finally:
+            # reset umask to initial value
+            os.umask(previous_umask)
```
gaborbernat: move comment into the finally line

src/filelock/_api.py
```diff
@@ -39,14 +39,20 @@ def __exit__(
...
+        mode: int = 0o644,
+    ) -> None:
         """
         Create a new lock object.
 
         :param lock_file: path to the file
         :param timeout: default timeout when acquiring the lock, in seconds. It will be used as fallback value in
         the acquire method, if no timeout value (``None``) is given. If you want to disable the timeout, set it
         to a negative value. A timeout of 0 means, that there is exactly one attempt to acquire the file lock.
+        : param multiuser: if True, make the lock file permissions rw/rw/rw for all users
```
gaborbernat: mode not multiuser now 👍

docs/changelog.rst
```diff
@@ -1,5 +1,8 @@
 Changelog
 =========
+v3.9.1 (2023-03-13)
```
gaborbernat: 3.10.0 as is feature

src/filelock/_api.py
```diff
@@ -52,7 +52,7 @@ def __init__(
         :param timeout: default timeout when acquiring the lock, in seconds. It will be used as fallback value in
         the acquire method, if no timeout value (``None``) is given. If you want to disable the timeout, set it
         to a negative value. A timeout of 0 means, that there is exactly one attempt to acquire the file lock.
-        : param multiuser: if True, make the lock file permissions rw/rw/rw for all users
+        : param mode: file permissions for the lockfile to have. default: 0o644 (-rw-r--r--)
```
gaborbernat: remove default, it's already in the header of the method

src/filelock/_api.py
```diff
@@ -165,6 +174,7 @@ def acquire(
         lock_id = id(self)
         lock_filename = self._lock_file
         start_time = time.monotonic()
+        previous_umask = os.umask(0)
```
gaborbernat: Shouldn't this try/catch be scopped just to _acquire?

jahrules: Looking at _acquire; it will never throw a BaseException error, so it should be safe to just wrap self._acquire() with a set and reset.

The try block is probably a candidate for future improvement because it's quite strangely written; but that seems outside of the scope of this PR.

gaborbernat (changes requested):
Can you rebase on main, and fix merge conflict?

src/filelock/_api.py
gaborbernat: What if you get a keyboardinterrupt; let's use try/finally please.

jahrules:
> 
rebased and fixed all conflicts.

gaborbernat (approved):

PR #194, opened 2023-02-06
use time.perf_counter instead of time.monotonic

zpz:
Use `time.perf_counter` instead of `time.monotonic` for calculating timeouts. The former has much higher clock resolution with comparable overhead, hence giving user finer timing control.

See
https://github.com/tox-dev/py-filelock/issues/191

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,9 @@
 Changelog
 =========
+in progress
+-----------
+- Use ``time.perf_counter`` instead of ``time.monotonic`` for calculating timeouts.
+
 v3.9.0 (2022-12-28)
 -------------------
 - Move build backend to ``hatchling`` :pr:`185 - by :user:`gaborbernat`.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,6 @@
 Changelog
 =========
-in progress
+v3.9.1 (2023-03-14)
 -----------
 - Use ``time.perf_counter`` instead of ``time.monotonic`` for calculating timeouts.
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,7 +1,7 @@
 Changelog
 =========
 v3.9.1 (2023-03-14)
------------
+-------------------
 - Use ``time.perf_counter`` instead of ``time.monotonic`` for calculating timeouts.
 
 v3.9.0 (2022-12-28)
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -1,5 +1,9 @@
 Changelog
 =========
+in progress
```
gaborbernat: 3.10 with todays date please

gaborbernat (changes requested):

docs/changelog.rst
gaborbernat: Also please ensure the CI passes.

zpz: > 3.10 with todays date please

I don't know your plan for the version number and date of the next release, so I don't think I should write those. Some ppl call this section "[Unreleased]", that's the same idea

https://keepachangelog.com/en/1.0.0/

zpz: > Also please ensure the CI passes.

Now flake8 check passes

src/filelock/_api.py
```diff
@@ -164,7 +164,7 @@ def acquire(
 
         lock_id = id(self)
         lock_filename = self._lock_file
-        start_time = time.monotonic()
+        start_time = time.perf_counter()  # noqa: SC200
```
gaborbernat: Remove these noqa

gaborbernat (changes requested):

src/filelock/_api.py
gaborbernat: Should add the word to https://github.com/tox-dev/py-filelock/blob/main/whitelist.txt#L3 instead

docs/changelog.rst
gaborbernat: Also just pick the next bugfix version instead of unreleased/in-progress

src/filelock/_api.py
zpz: what word?

docs/changelog.rst
zpz: sorry, it's your project. I don't want to set version number for you.

gaborbernat: I'm asking you to do it though 😉

src/filelock/_api.py
gaborbernat: What word you added this noqa for.

docs/changelog.rst
zpz: done

src/filelock/_api.py
zpz: done, I guess

gaborbernat (approved):

Issue #202, opened 2023-03-21
Error on Pickling Error Message: `Timeout.__init__() missing 1 required positional argument: 'lock_file'`


TheMatt2:
## Description

 If a `filelock.Timeout()` occurs, the resulting error does not pickle correctly.
This is a problem, because the module `concurrent.futures` assumes exceptions are pickled correctly, so they can be
sent back to the main process.

filelock: 3.10.0
Python: 3.11.2

## Steps to Reproduce

Run the following file
```python
# demo.py
import time
import concurrent.futures
from filelock import FileLock

file_path = "high_ground.txt"
lock_path = "high_ground.txt.lock"

lock = FileLock(lock_path, timeout=1)

def worker_use_lock(msg):
    with lock:
        with open(file_path, "a") as f:
            f.write(f"{msg}\n")

        # Simulate doing some work with the lock held
        time.sleep(10)
        return msg

def main():
    # Create executor to call function in another process
    with concurrent.futures.ProcessPoolExecutor(2) as executor:
        messages = ["Hello there!", "General Kenobi!"]
        for msg in executor.map(worker_use_lock, messages):
            print(f"Saved message {msg}")

if __name__ == "__main__":
    main()
```

It will fail with traceback:

```
concurrent.futures.process._RemoteTraceback: 
'''
Traceback (most recent call last):
  File ".../concurrent/futures/process.py", line 414, in wait_result_broken_or_wakeup
    result_item = result_reader.recv()
                  ^^^^^^^^^^^^^^^^^^^^
  File ".../multiprocessing/connection.py", line 250, in recv
    return _ForkingPickler.loads(buf.getbuffer())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Timeout.__init__() missing 1 required positional argument: 'lock_file'
'''

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "demo.py", line 27, in <module>
    main()
  File "demo.py", line 23, in main
    for msg in executor.map(worker_use_lock, messages):
  File ".../concurrent/futures/process.py", line 597, in _chain_from_iterable_of_lists
    for element in iterable:
  File ".../concurrent/futures/_base.py", line 619, in result_iterator
    yield _result_or_cancel(fs.pop())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../concurrent/futures/_base.py", line 317, in _result_or_cancel
    return fut.result(timeout)
[...]

PR #203, opened 2023-03-22
Properly pickle of Timeout objects + test cases

TheMatt2:
This is a fix for bug #202.

Timeout now registers a `__reduce__()` function to allow for proper pickling.
Tests check `Timeout` object has proper attributes and pickle successfully.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,8 +1,12 @@
 Changelog
 =========
+v3.10.1 (2023-03-22)
+--------------------
+- Handle pickle for :class:`filelock.Timeout` :pr:`203` - by :user:`TheMatt2`.
+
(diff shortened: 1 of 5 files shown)


TheMatt2:
```
type: commands[0]> mypy --strict src/filelock
src\filelock\_error.py:21: error: Returning Any from function declared to return "str"  [no-any-return]
src\filelock\_error.py:26: error: Returning Any from function declared to return "str"  [no-any-return]
```

Well... looks like I have some type checking to work out...

TheMatt2:
```
- hook id: flake8
- exit code: 1

tests/test_filelock.py:295:5: SC200 Possibly misspelt word: 'timeout2'
```

Well... ok then

gaborbernat (approved):

PR #206, opened 2023-03-22
changed from os.umask to os.chmod

jahrules:
addressing bug reported in https://github.com/tox-dev/py-filelock/issues/204

### DIFF
--- a/whitelist.txt
+++ b/whitelist.txt
@@ -4,7 +4,6 @@ autosectionlabel
 caplog
 eacces
 extlinks
-fchmod
 filelock
 filemode
 frameinfo
--- a/whitelist.txt
+++ b/whitelist.txt
@@ -4,6 +4,7 @@ autosectionlabel
 caplog
 eacces
 extlinks
+fchmod
 filelock
 filemode
 frameinfo
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,9 @@
 Changelog
 =========
+v3.10.2 (2023-03-22)
+--------------------
+- Bug fix for using filelock with threaded programs causing undesired file permissions - by :user:`jahrules`.
+
 v3.10.1 (2023-03-22)
 --------------------
 - Handle pickle for :class:`filelock.Timeout` :pr:`203` - by :user:`TheMatt2`.
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -179,11 +179,7 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
                 with self._thread_lock:
                     if not self.is_locked:
                         _LOGGER.debug("Attempting to acquire lock %s on %s", lock_id, lock_filename)
-                        previous_umask = os.umask(0)
-                        try:
-                            self._acquire()
-                        finally:
-                            os.umask(previous_umask)  # reset umask to initial value
+                        self._acquire()
                 if self.is_locked:
                     _LOGGER.debug("Lock %s acquired on %s", lock_id, lock_filename)
                     break
--- a/src/filelock/_soft.py
+++ b/src/filelock/_soft.py
@@ -21,7 +21,7 @@ class SoftFileLock(BaseFileLock):
             | os.O_TRUNC  # truncate the file to zero byte
         )
         try:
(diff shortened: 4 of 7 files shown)


gaborbernat (approved):

gaborbernat (approved):

whitelist.txt
```diff
@@ -4,6 +4,7 @@ autosectionlabel
 caplog
 eacces
 extlinks
+fchmod
```
gaborbernat: remove this?

jahrules: will do

gaborbernat (changes requested):
No merge commits please, use rebase instead.

jahrules:
> No merge commits please, use rebase instead.

apologies. I have fixed this.

here's what I've learned

Windows (annoyingly) doesn't suppose chmod. However, the default umask for windows allows -rw-rw-rw (vs UNIX's more restrictive default). So the following updates were made:

_unix.py
call os.chmod after opening the file to change the permissions

_windows.py preserve the os.open with the mode= option since the windows defaults will allow for -rw-rw-rw

_soft.py - remove the changed mode. Since the file is deleted after use, the SoftFileLock never needed to support -rw-rw-rw in the first place.

gaborbernat (approved):

xiormeesh:
Today after after this change poetry installs in Docker/Podman containers started to fail (it depends on filelock which was bumped from 3.10.0 to 3.10.12), from our CI log:

```
$ pip install --user poetry

Successfully installed <...> filelock-3.10.2 <...>

$ $HOME/.local/bin/poetry env use python3.8
Creating virtualenv <...>-YGs9DcwD-py3.8 in /home/workspace/.cache/pypoetry/virtualenvs
failed to build image pip, wheel, setuptools because:
  File "/home/workspace/.local/lib/python3.10/site-packages/virtualenv/seed/embed/via_app_data/via_app_data.py", line 52, in _install
    with parent.non_reentrant_lock_for_key(wheel_img.name):
  File "/usr/local/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/workspace/.local/lib/python3.10/site-packages/virtualenv/util/lock.py", line 146, in non_reentrant_lock_for_key
    with _CountedFileLock(str(self.path / f"{name}.lock")):
  File "/home/workspace/.local/lib/python3.10/site-packages/filelock/_api.py", line 227, in __enter__
    self.acquire()
  File "/home/workspace/.local/lib/python3.10/site-packages/virtualenv/util/lock.py", line 29, in acquire
    super().acquire(timeout, poll_interval)
  File "/home/workspace/.local/lib/python3.10/site-packages/filelock/_api.py", line 182, in acquire
    self._acquire()
  File "/home/workspace/.local/lib/python3.10/site-packages/filelock/_unix.py", line 36, in _acquire
    os.chmod(fd, self._mode)
PermissionError: [Errno 1] Operation not permitted: 3
```

The only w/a is after installing poetry forcefully downgrading filelock to <= 3.10.1

TheMatt2:
I think there is an issue with #206 

By calling `open()` without mode, and then following it with `os.chmod()` this leads to a situation where `os.chmod()` may set stricter permissions than `open()`. This means there is an opportunity between file creation and file permission set where less privileged users can influence the file. This is a problem.

```python
import os, filelock

# Program makes os.umask more permissive
os.umask(0)

with filelock.FileLock("/path/to/lock"):
    # Between lock creation and destruction, there is a moment where the permissions are rwxrwxrwx.
    ...
```

https://github.com/tox-dev/py-filelock/blob/b85b5cbb3db00bfc87d75b8551be341ea8696092/src/filelock/_unix.py#L33-L36
I think the permissions should be set by `os.open()` _and_ `os.chmod()` to avoid an issue.


Also, it may be a good idea have a way for the permissions of the lock file to be set by `os.umask()` should the user wish. Perhaps the default permissions for FileLock should be `None`, and signals that `chmod()` is not called and permissions are whatever `os.umask()` makes them to be.

Related Stack post: https://stackoverflow.com/questions/36745577/how-do-you-create-in-python-a-file-with-permissions-other-users-can-write#comment105490504_36745806

Issue #67, opened 2020-04-16
Trying to infinetely aquire a lock if flock is not implemented


PeterZhizhin:
I'm using a lustre file system. Looks like it doesn't support flock syscall.

Here is a small script I'm trying to run:
```python
import filelock

t = filelock.FileLock('test.lock')
with t:
  pass
```

Here is the output:
```
$ python test.py
... (Runs indefinitely)
```

Let's try running under strace:
```
$ strace python test.py
<A lot of garbage>
select(0, NULL, NULL, NULL, {0, 50000}) = 0 (Timeout)
open("test.lock", O_RDWR|O_CREAT|O_TRUNC|O_CLOEXEC, 0777) = 3
flock(3, LOCK_EX|LOCK_NB)               = -1 ENOSYS (Function not implemented)
close(3)                                = 0
select(0, NULL, NULL, NULL, {0, 50000}) = 0 (Timeout)
open("test.lock", O_RDWR|O_CREAT|O_TRUNC|O_CLOEXEC, 0777) = 3
flock(3, LOCK_EX|LOCK_NB)               = -1 ENOSYS (Function not implemented)
close(3)                                = 0
select(0, NULL, NULL, NULL, {0, 50000}) = 0 (Timeout)
open("test.lock", O_RDWR|O_CREAT|O_TRUNC|O_CLOEXEC, 0777) = 3
flock(3, LOCK_EX|LOCK_NB)               = -1 ENOSYS (Function not implemented)
close(3)                                = 0
select(0, NULL, NULL, NULL, {0, 50000}) = 0 (Timeout)
open("test.lock", O_RDWR|O_CREAT|O_TRUNC|O_CLOEXEC, 0777) = 3
flock(3, LOCK_EX|LOCK_NB)               = -1 ENOSYS (Function not implemented)
close(3)                                = 0
<And so on>
```

Solution: inside `UnixFileLock` do not ignore all exceptions. If OSError is raised with errno=ENOSYS, then show a warning, delete the lock file and fallback to SoftFileLock.

jarednielsen:
Also hitting this issue. Any resolution?

gaborbernat:
Hello, if you make a PR for this to fix it (with tests) we would be happy to review it, thanks!

PR #212, opened 2023-03-25
proposed fix for issue #67

jahrules:
This is a proposed fix for https://github.com/tox-dev/py-filelock/issues/67

As suggested in the issue; we're checking the OSError for NotImplemented (which is errno 38).

OSError 38 is not fatal; so this results in the reported issue of an indeterminate loop.

I don't see a clear way to "fall back" from a UnixFileLock to a SoftFileLock; so my proposal is to raise a fatal error with a clear message to the end user.

I also don't see a good way to create a testcase for this; as raising an OSError code 38 requires a condition that can't be easily simulated. I'm open to ideas for this if I'm missing something obvious.

### DIFF
--- a/src/filelock/_unix.py
+++ b/src/filelock/_unix.py
@@ -39,8 +39,10 @@ else:  # pragma: win32 no cover
                 pass  # This locked is not owned by this UID
             try:
                 fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-            except OSError:
+            except OSError as err:
                 os.close(fd)
+                if err.errno == 38:  # NotImplemented error number
+                    raise NotImplementedError("FileSystem does not appear to support flock; user SoftFileLock instead")
             else:
                 self._lock_file_fd = fd
 
--- a/src/filelock/_unix.py
+++ b/src/filelock/_unix.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 import sys
+from errno import ENOSYS
 from typing import cast
 
 from ._api import BaseFileLock
@@ -39,9 +40,9 @@ else:  # pragma: win32 no cover
                 pass  # This locked is not owned by this UID
             try:
                 fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-            except OSError as err:
+            except OSError as exception:
                 os.close(fd)
-                if err.errno == 38:  # NotImplemented error number
+                if exception.errno == ENOSYS:  # NotImplemented error
                     raise NotImplementedError("FileSystem does not appear to support flock; user SoftFileLock instead")
             else:
                 self._lock_file_fd = fd
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,10 @@ repos:
     hooks:
       - id: tox-ini-fmt
         args: ["-p", "fix"]
+  - repo: https://github.com/tox-dev/pyproject-fmt
+    rev: "0.9.2"
+    hooks:
+      - id: pyproject-fmt
   - repo: https://github.com/PyCQA/flake8
     rev: 6.0.0
     hooks:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,9 @@
 Changelog
 =========
+v3.10.5 (2023-03-25)
+--------------------
+- Add explicit error check as certain UNIX filesystems do not support flock. by :user:`jahrules`.
+
 v3.10.4 (2023-03-24)
 --------------------
 - Update os.open to preserve mode= for certain edge cases. by :user:`jahrules`.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ optional-dependencies.docs = [
 optional-dependencies.testing = [
   "covdefaults>=2.3",
   "coverage>=7.2.2",
+  "diff-cover>=7.5",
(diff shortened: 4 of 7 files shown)


src/filelock/_unix.py
```diff
@@ -39,8 +39,10 @@ def _acquire(self) -> None:
                 pass  # This locked is not owned by this UID
             try:
                 fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-            except OSError:
+            except OSError as err:
                 os.close(fd)
+                if err.errno == 38:  # NotImplemented error number
```
TheMatt2: Please use `err.errno == errno.ENOSYS` instead of hard coding 38.

On common _linux_ it looks like ENOSYS is 38. On _MacOS_ [ENOSYS is 78](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/intro.2.html).

TheMatt2:
>I also don't see a good way to create a testcase for this; as raising an OSError code 38 requires a condition that can't be easily simulated.

I think there is actually a relatively straight forward way to test this condition.
You can stub `fcntl.flock` to test the function.

(This is psuedo code for it, I haven't tested it)
```python
import errno
import fcntl
import filelock

def dummy_flock(fd, operation):
    raise OSError(errno.ENOSYS, "flock not supported")

def test_flock_not_implemented():
    _fcntl_flock = fcntl.flock
    try:
        fcntl.flock = dummy_flock
        with pytest.raises(NotImplementedError):
            with filelock.FileLock("/path/to/test/file"): pass
    finally:
        fcntl.flock = _fcntl_flock
```

gaborbernat (changes requested):
Can you try adding the suggested test and changelog entry?

jahrules:
> > I also don't see a good way to create a testcase for this; as raising an OSError code 38 requires a condition that can't be easily simulated.
> 
> I think there is actually a relatively straight forward way to test this condition. You can stub `fcntl.flock` to test the function.
> 
> (This is psuedo code for it, I haven't tested it)
> 
> ```python
> import errno
> import fcntl
> import filelock
> 
> def dummy_flock(fd, operation):
>     raise OSError(errno.ENOSYS)
> 
> def test_flock_not_implemented():
>     _fcntl_flock = fcntl.flock
>     try:
>         fcntl.flock = dummy_flock
>         with pytest.raises(NotImplementedError):
>             with filelock.FileLock("/path/to/test/file"): pass
>     finally:
>         fcntl.flock = _fcntl_flock
> ```

This worked very well thanks.

I write a lot of functional and regression type tests, so my unit-test-foo is a bit rusty.

jahrules:
After a lot of fighting with the Unix type checker I finally got it to pass. (frustration)
the Windows type checker is still failing because it sees the flock imports even though the testcase is only set to run for Unix. 

I'm not sure how to tell the windows type checker to ignore these...

src/filelock/_unix.py
```diff
@@ -39,8 +40,10 @@ def _acquire(self) -> None:
                 pass  # This locked is not owned by this UID
             try:
                 fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
-            except OSError:
+            except OSError as exception:
                 os.close(fd)
+                if exception.errno == ENOSYS:  # NotImplemented error
+                    raise NotImplementedError("FileSystem does not appear to support flock; user SoftFileLock instead")
```
TheMatt2, line 46: typo fix.
Perhaps `"Filesystem does not support file locking. Please use SoftFileLock instead."`?

TheMatt2:
> I'm not sure how to tell the windows type checker to ignore these...

It seems the way to get the type checker to ignore these is to change 

```python
@pytest.mark.skipif(sys.platform == "win32", reason="Windows filesystems do not support flock")
def test_flock_not_implemented_unix(tmp_path: Path) -> None:
```

to

```python
def test_flock_not_implemented_unix(tmp_path: Path) -> None:
    if sys.platform == "win32":
        pytest.skip("Windows filesystems do not support flock")
```

Will make the type checker happy on windows. That way you don't need to have ignore statements that Ubuntu is now complaining about!

(I must admit, Github Copilot actually gets the credit for figuring this out, which is just a little scary)

jahrules:
> > I'm not sure how to tell the windows type checker to ignore these...
> 
> It seems the way to get the type checker to ignore these is to change
> 
> ```python
> @pytest.mark.skipif(sys.platform == "win32", reason="Windows filesystems do not support flock")
> def test_flock_not_implemented_unix(tmp_path: Path) -> None:
> ```
> 
> to
> 
> ```python
> def test_flock_not_implemented_unix(tmp_path: Path) -> None:
>     if sys.platform == "win32":
>         pytest.skip("Windows filesystems do not support flock")
> ```
> 
> Will make the type checker happy on windows. That way you don't need to have ignore statements that Ubuntu is now complaining about!
> 
> (I must admit, Github Copilot actually gets the credit for figuring this out, which is just a little scary)

too bad mypy couldnt just detect the skipif decorator. Talking to some people; apparently in certain circumstances it can... idk. Im just happy it works now.

gaborbernat (approved):

gaborbernat:
> too bad mypy couldnt just detect the skipif decorator. Talking to some people; apparently in certain circumstances it can... idk. Im just happy it works now.

Check my feedback fix, that avoids the issue.

src/filelock/_unix.py
TheMatt2: fixed

PR #213, opened 2023-03-25
Bugfix/147

### LINKED ISSUE #147: What happens if another type of OSError is raised when attempting to create a soft lock?
I ran into a strange bug when trying to lock a file on a network file-system mounted inside a container, where the lock file was created but for some reason it seems as though the file-handle wasn't properly returned. My process then got stuck waiting for the lock to be released (when it had in fact created the lock). Looking at the following code, it seems that if the OSError errno isn't EEXIST, ENOENT or EACCES, then it is assumed the file is locked

https://github.com/tox-dev/py-filelock/blob/4730a40b87cc4b094330b2af7723658428323d60/src/filelock/_soft.py#L23-L32

wouldn't it be more robust to do something like

```python
 try: 
     fd = os.open(self._lock_file, mode) 
 except OSError as exception: 
     if (exception.errno == EEXIST # expected if cannot lock 
             or (if exception.errno == EACCES and sys.platform == "win32"):  # note windows does not allow you to make a folder r/o only files 
         pass 
     else:
         raise
```

Or do you actually want the code to keep attempting to try creating the lock on other OSErrors?

jahrules:
Resolves https://github.com/tox-dev/py-filelock/issues/147

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,5 +1,9 @@
 Changelog
 =========
+v3.10.6 (2023-03-25)
+--------------------
+- Enhance the robustness of the try/catch block in _soft.py. by :user: `jahrules`.
+
 v3.10.5 (2023-03-25)
 --------------------
 - Add explicit error check as certain UNIX filesystems do not support flock. by :user:`jahrules`.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -2,7 +2,7 @@ Changelog
 =========
 v3.10.6 (2023-03-25)
 --------------------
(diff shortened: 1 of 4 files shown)


gaborbernat (changes requested):
Please rebase on top of main.

gaborbernat (changes requested):
Test please 🥺

jahrules:
> Test please pleading_face

The mocker module that you found was very helpful for this. Thanks.

gaborbernat (approved):

Issue #225, opened 2023-04-06
document generation failing


gaborbernat:
See https://readthedocs.org/projects/py-filelock/builds/20040292/

@csm10495 can you check seems this slipped through the cracks with your PR

csm10495:
https://github.com/tox-dev/py-filelock/pull/226

PR #226, opened 2023-04-06
Fix: #225 :Get rid of warning about inability to link to _thread._local.

csm10495:
I'm not sure why it resolves this way, but there is no link for that pathing

Related sphinx issue: https://github.com/sphinx-doc/sphinx/issues/9338

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -2,6 +2,13 @@ from __future__ import annotations
 
 from datetime import date, datetime
 
+from docutils.nodes import Element, Text
+from sphinx.addnodes import pending_xref
+from sphinx.application import Sphinx
+from sphinx.builders import Builder
+from sphinx.domains.python import PythonDomain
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -24,7 +24,7 @@
 
 intersphinx_mapping = {"python": ("https://docs.python.org/3", None)}
 nitpicky = True
-nitpick_ignore = []
+nitpick_ignore = [("py:class", "_thread._local")]
```
gaborbernat: Instead of ignore can we map it to whatever stdlib resolves?

csm10495: I'd love to, but I couldn't seem to figure out how to get it to understand that it's threading.local (and not the internal implementation detail version).

My sphinx isn't the greatest

csm10495:
That's quite the impressive workaround @gaborbernat 👍

gaborbernat (approved):

PR #231, opened 2023-04-17
Fix lock hang on Windows

TheMatt2:
Solves an issue where `FileLock` and `SoftFileLock` could hang sit in an infinite loop if `PermissionError` or `OSError` is encountered on Windows while trying to lock the file.

Originally mentioned on #147.

On Windows, if the name of the file given is invalid, such as forgetting to to double backspace on windows, the EINVAL message is not accounted for in the error handling.

```python
import filelock

filename = "C:\path\to\lock\but\backslash\is\not\escaped"
with filelock.FileLock(filename):
    pass # waits forever and never completes
```

If you had just tried to open the file, you would see the following message:

```
filename = "C:\path\to\lock\but\backslash\is\not\escaped"
f = open(filename)
```

```
  File "<stdin>", line 1, in <module>
OSError: [Errno 22] Invalid argument: 'C:\\path\to\\lock\x08ut\x08ackslash\\is\not\\escaped'
```

The issue that is being danced around is it seems `PermissionError` will randomly occur on `os.open()` when trying to open the lock file. I have been unable to figure out *why* Windows raises `PermissionError`, but that seems somewhat a mute point.

https://github.com/tox-dev/py-filelock/blob/dceb9e5b7e4edd232435521f20efe189fe88f71b/src/filelock/_windows.py#L24-L28

The problem is any error raised, other than the one explicitly handled, ignoring other errors.
This is corrected by ignoring some explicit errors, but allowing other errors to be raised.

This also adds a check for if the path of the lock file is a directory. On Windows, trying to open a path that is a directory as a file would raise a `PermissionError` that can't be differentiated from the exception that is sometimes raised from just trying to open the file in the middle of a bunch of locking and unlocking.

Renamed `utils.raise_on_exist_ro_file(filename)` to `utils.raise_unwritable_file(filename)` to better describe what is does, now that the directory check has been added to it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,7 +54,7 @@ repos:
       - id: flake8
         additional_dependencies:
           - flake8-bugbear==23.3.23
-          - flake8-comprehensions==3.11.1
+          - flake8-comprehensions==3.12
           - flake8-pytest-style==1.7.2
           - flake8-spellcheck==0.28
           - flake8-unused-arguments==0.0.13
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,13 +37,13 @@ dynamic = [
 optional-dependencies.docs = [
   "furo>=2023.3.27",
   "sphinx>=6.1.3",
-  "sphinx-autodoc-typehints!=1.23.4,>=1.22",
+  "sphinx-autodoc-typehints!=1.23.4,>=1.23",
 ]
 optional-dependencies.testing = [
   "covdefaults>=2.3",
   "coverage>=7.2.3",
   "diff-cover>=7.5",
-  "pytest>=7.2.2",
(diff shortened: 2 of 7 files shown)


TheMatt2:
Yes, flake. "unwritable" is a word...

TheMatt2:
I'm actually not sure why I'm getting the flake8 error

```
tests/test_filelock.py:137:2: PT006 wrong name(s) type in @pytest.mark.parametrize, expected tuple
```

~~Anyone have suggestions?~~ Found it

https://github.com/m-burst/flake8-pytest-style/blob/master/docs/rules/PT006.md

Should be formatted as a tuple... which I didn't know `pytest` supported. Go figure.

gaborbernat (approved):

PR #232, opened 2023-04-18
Conditionally disable/enable thread-local lock behavior.

### LINKED ISSUE #230: filelock 3.11 breaks our multi-threaded parallel processing use case
Hi,

In #219 @csm10495 wondered
> I'm curious though: Is there ever a case where this would break someone? I guess if someone expected the object to be the same across threads.. but then why would they be using the lock to begin with?

I guess we have a use case that breaks:

We have a multi-threaded processing pipeline that is set up roughly like this:
1. Initialize and create a `SoftFileLock` for the output file.
2. Prepare some data for step 3.
3. Generate the output file with the data from step 2.
4. Clean up and release the lock acquired in 1.

These steps are executed for multiple inputs, each generating a different output file. Step 2 for one input can run in parallel to step 3 for another input. However, step 3 for one input must not run in parallel to step 3 for another input, as it uses resources that shall not be shared.

The `SoftFileLock` is required because there may be another process that is executed on an overlapping set of input files (thus generating an overlapping set of output files). With filelock 3.10, we could pass the lock from the thread executing step 1 to the thread executing step 4, but this broke with #219.

For this use case, I don't see anything wrong with sharing the lock between threads, because the lock is intended to prevent concurrent access between different processes. What do you think?

csm10495:
Should resolve https://github.com/tox-dev/py-filelock/issues/230.

TLDR: We now have a thread_local=True by default. If set to False, locks will be re-entrant across threads (a-la 3.10.7 and older behavior).

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -6,6 +6,7 @@ import os
 import time
 import warnings
 from abc import ABC, abstractmethod
+from dataclasses import dataclass
 from threading import local
 from types import TracebackType
 from typing import Any
@@ -36,7 +37,42 @@ class AcquireReturnProxy:
         self.lock.release()
 
 
-class BaseFileLock(ABC, contextlib.ContextDecorator, local):
+@dataclass
+class FileLockContext:
+    """
+    A dataclass which holds the context for a ``BaseFileLock`` object.
+
+    The context is held in a separate class to allow optional use of thread local storage
+    via the ``ThreadLocalFileContext`` class.
+    """
+
+    # The path to the lock file.
+    lock_file: str
+
+    # The default timeout value.
+    timeout: float
+
+    # The mode for the lock files
+    mode: int
+
+    # The file descriptor for the *_lock_file* as it is returned by the os.open() function.
+    # This file lock is only NOT None, if the object currently holds the lock.
+    lock_file_fd: int | None = None
+
+    # The lock counter is used for implementing the nested locking mechanism. Whenever the lock is acquired, the
+    # counter is increased and the lock is only released, when this value is 0 again.
+    lock_counter: int = 0
+
+
+class ThreadLocalFileContext(FileLockContext, local):
+    """
+    A thread local version of the ``FileLockContext`` class.
+    """
+
+    pass
+
+
+class BaseFileLock(ABC, contextlib.ContextDecorator):
     """Abstract base class for a file lock object."""
 
     def __init__(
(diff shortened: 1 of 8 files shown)


tests/test_filelock.py
```diff
@@ -576,3 +576,20 @@ def mess_with_file() -> None:
...
+
+@pytest.mark.parametrize("lock_type", [FileLock, SoftFileLock])
+def test_lock_can_be_non_thread_local(
+    tmp_path: Path,
+    lock_type: type[BaseFileLock],
+) -> None:
+    lock = lock_type(tmp_path / "test.lock", thread_local=False)
+
+    for _ in range(2):
+        t = threading.Thread(target=lock.acquire, kwargs={"timeout": 2})
```
gaborbernat: Don't use one letter variables outside of  comprehension please.

gaborbernat:
Also the CI is failing.

tests/test_filelock.py
csm10495: fixed

csm10495:
you've caught me in between pushes. Check again in a moment.

csm10495:
@gaborbernat now we're passing. Feel free to take a look. Thanks.

src/filelock/_api.py
```diff
@@ -52,29 +89,30 @@ def __init__(
...
-        # The default timeout value.
-        self._timeout: float = timeout
-
-        # The mode for the lock files
-        self._mode: int = mode
-
-        # The lock counter is used for implementing the nested locking mechanism. Whenever the lock is acquired, the
-        # counter is increased and the lock is only released, when this value is 0 again.
-        self._lock_counter: int = 0
+        # Whether or not this object is thread local or not
```
slanzmich:
```suggestion
        # Whether this object is thread local or not
```

src/filelock/_api.py
```diff
@@ -52,29 +89,30 @@ def __init__(
...
+        self.thread_local = thread_local
+
+        # Create the context. Note that external code should not work with the context directly
+        # and should instead use properties of this class.
+        context_kwargs: dict[str, Any] = {
+            "lock_file": os.fspath(lock_file),
+            "timeout": timeout,
+            "mode": mode,
+        }
+        self._context: FileLockContext | ThreadLocalFileContext
```
slanzmich:
```suggestion
        self._context: FileLockContext
```
should be enough, because `ThreadLocalFileContext` is also a `FileLockContext`.

I think `mypy` complained because the "first" branch assigned an instance of the subclass, and may not have complained if the `if ... else` branches would have been swapped (not suggesting that, though).

gaborbernat (approved):

PR #266, opened 2023-08-25
fixing narrow type-hint

### LINKED ISSUE #258: The type annotation for the `lock_file` `__init__` parameter is too narrow
`BaseFileLock.__init__` annotates `lock_file` as `str | os.PathLike[Any]`.  However, this excludes `bytes` which are still the basic representation of paths on most POSIX systems.  The implementation goes on to pass `lock_file` to `os.fspath` which both accepts and may return `bytes` values so there doesn't seem to be any reason to exclude `bytes` from the annotation.

keller00:
fixes #258


gaborbernat (changes requested):
test please

gaborbernat (approved):

src/filelock/_api.py
```diff
@@ -95,11 +95,13 @@ def __init__(
         If this is set to ``False`` then the lock will be reentrant across threads.
         """
         self._is_thread_local = thread_local
+        path = os.fspath(lock_file)
+        path_str = path.decode() if isinstance(path, bytes) else path
```
gaborbernat, line 99: Hmm, with this I'm no longer sure this is something we want to support. The user can put their casting at call side if they must, no?

keller00: Other parts of the code depend in this being a `str`, like these [debug logging lines](https://github.com/tox-dev/filelock/blob/09ef4128981559ea7a2c4f2c48937d48debd5e63/src/filelock/_api.py#L225) (otherwise we get `b'log lines'`).

So the choice for me was whether to do `filename.decode() if isinstance(filename, bytes) else filename` in multiple spots, or make sure that it's `str` at init time.

keller00:
@gaborbernat I don't mind if you decide to not accept the PR. I was trying to help close the low hanging issues. 😄 
The one thing that I'd like to change is this type-hint [here](https://github.com/tox-dev/filelock/blob/09ef4128981559ea7a2c4f2c48937d48debd5e63/src/filelock/_api.py#L81).
I believe `lock_file: str | os.PathLike[Any],` should instead be `lock_file: str | os.PathLike[AnyStr],`, but I'm happy to open a differentPR for that

gaborbernat:
> @gaborbernat I don't mind if you decide to not accept the PR. I was trying to help close the low hanging issues. 😄 
> The one thing that I'd like to change is this type-hint [here](https://github.com/tox-dev/filelock/blob/09ef4128981559ea7a2c4f2c48937d48debd5e63/src/filelock/_api.py#L81).
> I believe `lock_file: str | os.PathLike[Any],` should instead be `lock_file: str | os.PathLike[AnyStr],`, but I'm happy to open a differentPR for that

Let's do that 👍

Issue #78, opened 2020-09-21
Request for Asyncio Support


oldcai:
It would block other coroutines when acquiring locks now. 

Hope it can use unblocking ways to wait for the lock. Such as `asyncio.sleep`

By the way, in the meantime, I'm using this way to keep other coroutines running.

```
    acquired_lock = False
    while not acquired_lock:
        try:
            BEFORE_NET_START_LOCK.acquire(timeout=0)
            acquired_lock = True
        except Timeout:
            await asyncio.sleep(1)
```

Hope this code can help with someone facing the same problem.

xkortex:
This ought to be implemented with `async with`: https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with

This ought to improve performance as well, since it looks like acquire currently uses a busy-wait: https://github.com/benediktschmitt/py-filelock/blob/master/filelock.py#L284

gaborbernat:
> This ought to be implemented with `async with`: [python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with](https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with)

That would imply an event loop running, which is often not the case. So we cannot add asyncio support straight up, however if someone would make this library sansio type we'd be happy to review that PR :+1:

maxfischer2781:
I've gone over the code and as far as I can tell there is no way to make things *nicely* `async`.

Fundamentally, the locking via `_acquire` requires polling and all that `asyncio`/`trio`/... can do is provide an ``await sleep(poll_interval)`` instead of the current ``time.sleep(poll_interval)``. That is really bad for the event loop at the default `poll_interval=0.05`; some exponential backoff might help here.  
Alternatively, one could reimplement a blocking acquire and use a thread ([as suggested on SO](https://stackoverflow.com/a/64901794/5349916)). That is ideal if we definitely want to acquire a few locks; it is fatal with timeouts/cancellation and many locks, since the threads will keep on living "forever" in the background in the worst case.

-------------

*If* you are fine with having a busy loop, I would suggest (and can provide) an ``async with`` context manager that basically uses the ``while True:`` + ``await sleep`` to poll the lock. There is no way to make it truly sansio, but the `sleep` function could be parametrized easily.

This could either be a separate class (say ``AsyncFileLock``) that wraps a regular `FileLock` or merely adding the required magic methods for `async with` directly to `FileLock`. Would that be simple enough to add to the library?

[...]

PR #279, opened 2023-10-05
WIP: Async support

rohitsanj:
## Introduce async support to `filelock`

With this, you can now use filelock with your friendly neighborhood async event loop implementations.

### Async context manager usage
```python
from filelock.asyncio import FileLock

async def safe_write():
    lock = FileLock("foo.txt.lock")
    async with lock:
        with open("foo.txt", "w") as f:
            f.write("bar")
```

### Explicit `acquire`/`release` usage

```python
from filelock.asyncio import FileLock

async def safe_write():
    lock = FileLock("foo.txt.lock", timeout=4.2)
    try:
        await lock.acquire()
        with open("foo.txt", "w") as f:
            f.write("bar")
    finally:
        lock.release()
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ dist
 /pip-wheel-metadata
 /src/filelock/version.py
 venv*
+.venv*
 .python-version
 test.lock
 test.softlock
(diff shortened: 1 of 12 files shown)


gaborbernat (changes requested):
As I aluded to in https://github.com/tox-dev/filelock/issues/78#issuecomment-927655481 the only way I'd accept a such PR if you'd make the existing one sansio (https://snarky.ca/network-protocols-sans-i-o/), I do not want it to be a parallel implementation as you did here. If we'd go down that path I'd prefer to make it a separate library.

gaborbernat:
Stalled.

Issue #282, opened 2023-10-25
Bug: Reentrant behavior does not work with distinct lock objects on the same thread


nefrob:
#### Setup

- Python 3.11.3
- filelock 3.12.4
- MacOS

#### Actual:

```python
➜ python
>>> from filelock import FileLock
>>> lock_1 = FileLock("test.lock")
>>> lock_2 = FileLock("test.lock")
>>> lock_1.acquire()
<filelock._api.AcquireReturnProxy object at 0x100baa550>
>>> lock_1._context
ThreadLocalFileContext(lock_file='test.lock', timeout=-1, mode=420, lock_file_fd=3, lock_counter=1)
>>> lock_2._context
ThreadLocalFileContext(lock_file='test.lock', timeout=-1, mode=420, lock_file_fd=None, lock_counter=0)
>>> lock_2.acquire()
... hangs
```

Also it does not appear that releasing the lock removes the lock file (not sure if this is expected):

```python
>>> from filelock import FileLock
>>> lock = FileLock("test.lock")
>>> lock.acquire()
<filelock._api.AcquireReturnProxy object at 0x102ad80d0>
>>> lock.release()
```

```bash
➜ ls
test.lock
```

#### Expectation:

- `lock_2` can be acquired since it is locking the same file on the same thread per reentrancy definition in docs: 

> The lock objects are recursive locks, which means that once acquired, they will not block on successive lock requests

(https://py-filelock.readthedocs.io/en/latest/#tutorial)

gaborbernat:
PR welcome 👍

nefrob:
Closing this re #283 being merged/released.

hmaarrfk:
I'm not sure why this is the expected behavior.

The lock objects `lock_1` and `lock_2` are just two different objects.

I personally feel like #283 is a step backward.

gaborbernat:
It's optional opt in behavior so shouldn't cause you issues 🤔

PR #283, opened 2023-10-25
Support reentrant locking on lock file path via optional singleton instance

nefrob:
#### Description

Re: #282.

- When `_singleton_per_lock_file` is set instantiating a new lock with the same lock path will return the existing instance of the lock. This means you can reacquire the lock without passing the lock object around
- Otherwise a new lock object is returned every time (default behavior for backwards compatibility)

Based on the `flock` docs (and likely similar for windows/soft cases)

> If a process uses [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) (or similar) to obtain more than one
       file descriptor for the same file, these file descriptors are
       treated independently by flock().  An attempt to lock the file
       using one of these file descriptors may be denied by a lock that
       the calling process has already placed via another file
       descriptor.

(https://man7.org/linux/man-pages/man2/flock.2.html)

a Python solution seems most general and easy here. If anyone has time to dig into different OS file locking mechanisms and see if there is a native solution that would likely be preferable.

#### Action Items

- [x] docs updates?
- [ ] concurrency? If the use case fro non-thread-local locks is to create the lock on the main thread, share it with threads, then cleanup the lock when they complete then this should be safe as the threads will only ever read the instance from the class instances dictionary. If supporting hand-off of the lock to thread processes is within the thread-local options then we should acquire a global lock around the instances dictionary for safety (at the cost of speed)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,26 +6,26 @@ This page lists the steps needed to set up a development environment and contrib
 
 2. Install initial dependencies in a virtual environment:
 
-    ```shell
-    python -m venv venv
-    source venv/bin/activate
-    python -m pip install --upgrade pip 'tox>=4.2'
-    ```
+   ```shell
+   python -m venv venv
+   source venv/bin/activate
+   python -m pip install --upgrade pip 'tox>=4.2'
+   ```
 
 3. Run tests
 
-    ```shell
-    tox run
-    ```
+   ```shell
+   tox run
+   ```
 
-    or for a specific python version
+   or for a specific python version
 
-    ```shell
-    tox run -f py311
-    ```
+   ```shell
+   tox run -f py311
+   ```
 
 4. Running other tox commands (ex. linting)
 
-    ```shell
-    tox -e fix
-    ```
+   ```shell
+   tox -e fix
+   ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,31 @@
+# Contributing
+
+This page lists the steps needed to set up a development environment and contribute to the project.
+
+1. Fork and clone this repo
+
+2. Install initial dependencies in a virtual environment:
+
+    ```shell
+    python -m venv venv
+    source venv/bin/activate
+    python -m pip install --upgrade pip 'tox>=4.2'
+    ```
+
+3. Run tests
+
+    ```shell
+    tox run
+    ```
+
+    or for a specific python version
+
+    ```shell
+    tox run -f py311
+    ```
+
+4. Running other tox commands (ex. linting)
+
+    ```shell
+    tox -e fix
+    ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,11 +2,11 @@
 
 This page lists the steps needed to set up a development environment and contribute to the project.
 
-1. Fork and clone this repo
+1. Fork and clone this repo.
 
-2. Install `tox` via `pipx`: https://tox.wiki/en/4.11.3/installation.html#via-pipx
+2. [Install tox](https://tox.wiki/en/latest/installation.html#via-pipx).
 
-3. Run tests
+3. Run tests:
 
    ```shell
    tox run
@@ -18,7 +18,7 @@ This page lists the steps needed to set up a development environment and contrib
    tox run -f py311
    ```
 
-4. Running other tox commands (ex. linting)
+4. Running other tox commands (ex. linting):
 
    ```shell
    tox -e fix
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,13 +4,7 @@ This page lists the steps needed to set up a development environment and contrib
 
 1. Fork and clone this repo
 
-2. Install initial dependencies in a virtual environment:
-
-   ```shell
-   python -m venv venv
-   source venv/bin/activate
-   python -m pip install --upgrade pip 'tox>=4.2'
-   ```
+2. Install `tox` via `pipx`: https://tox.wiki/en/4.11.3/installation.html#via-pipx
 
 3. Run tests
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ This page lists the steps needed to set up a development environment and contrib
    tox run -f py311
    ```
 
-4. Running other tox commands (ex. linting):
+4. Running other tox commands (eg. linting):
 
    ```shell
    tox -e fix
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -9,6 +9,7 @@ from abc import ABC, ABCMeta, abstractmethod
 from dataclasses import dataclass
 from threading import local
 from typing import TYPE_CHECKING, Any, ClassVar
+from weakref import WeakValueDictionary
 
 from ._error import Timeout
 
@@ -80,15 +81,18 @@ class SingletonPerLockFileMeta(ABCMeta):
     lock (when specified).
     """
 
-    _instances: ClassVar[dict[str, object]] = {}
+    _instances: ClassVar[WeakValueDictionary] = WeakValueDictionary()
 
-    def __call__(cls, lock_file: str, *args: Sequence, **kwargs: Mapping) -> object:
+    def __call__(cls, lock_file: str | os.PathLike[str], *args: Sequence, **kwargs: Mapping) -> object:
         if not kwargs.get("singleton_per_lock_file", False):
             return super().__call__(lock_file, *args, **kwargs)
 
-        if lock_file not in cls._instances:
-            cls._instances[lock_file] = super().__call__(lock_file, *args, **kwargs)
-        return cls._instances[lock_file]
+        instance = cls._instances.get(str(lock_file))
+        if not instance:
+            instance = super().__call__(lock_file, *args, **kwargs)
+            cls._instances[str(lock_file)] = instance
+
+        return instance
 
 
 class BaseFileLock(ABC, contextlib.ContextDecorator, metaclass=SingletonPerLockFileMeta):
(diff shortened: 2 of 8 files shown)


src/filelock/_api.py
```diff
@@ -73,15 +75,36 @@ class ThreadLocalFileContext(FileLockContext, local):
     """A thread local version of the ``FileLockContext`` class."""
 
 
-class BaseFileLock(ABC, contextlib.ContextDecorator):
+class SingletonPerLockFileMeta(ABCMeta):
```
nefrob: We inherit from `ABCMeta` here rather than `type` since `BaseFileLock` inherits from `ABC` and we would otherwise have a metaclass conflict.

src/filelock/_api.py
```diff
@@ -73,15 +75,36 @@ class ThreadLocalFileContext(FileLockContext, local):
...
     """Abstract base class for a file lock object."""
 
-    def __init__(
+    def __init__(  # noqa: PLR0913
         self,
         lock_file: str | os.PathLike[str],
         timeout: float = -1,
         mode: int = 0o644,
         thread_local: bool = True,  # noqa: FBT001, FBT002
+        singleton_per_lock_file: bool = False,  # noqa: FBT001, FBT002
```
nefrob: Naming suggestions welcome.

src/filelock/_api.py
```diff
@@ -73,15 +75,36 @@ class ThreadLocalFileContext(FileLockContext, local):
...
-class BaseFileLock(ABC, contextlib.ContextDecorator):
+class SingletonPerLockFileMeta(ABCMeta):
+    """
+    Metaclass for the ``BaseFileLock`` class which ensures that only one instance of the class is created per
+    lock (when specified).
+    """
+
+    _instances: ClassVar[WeakValueDictionary] = WeakValueDictionary()
+
+    def __call__(cls, lock_file: str | os.PathLike[str], *args: Sequence, **kwargs: Mapping) -> object:
```
nefrob: Hmm just realizing the same behavior can be achieved via `__new__` instead of creating a metaclass. I'm going to refactor the approach since that seems clearer.

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+This page lists the steps needed to set up a development environment and contribute to the project.
+
+1. Fork and clone this repo
+
+2. Install initial dependencies in a virtual environment:
+
+   ```shell
+   python -m venv venv
+   source venv/bin/activate
+   python -m pip install --upgrade pip 'tox>=4.2'
```
gaborbernat: Just link to https://tox.wiki/en/4.11.3/installation.html#via-pipx, don't duplicate documentation.

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
+# Contributing
+
+This page lists the steps needed to set up a development environment and contribute to the project.
+
+1. Fork and clone this repo
+
+2. Install `tox` via `pipx`: https://tox.wiki/en/4.11.3/installation.html#via-pipx
```
gaborbernat: Use latest instead of concrete version, also make it a markdown link, not explicit. Something like:

[Install tox](https://tox.wiki/en/latest/installation.html#via-pipx).

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
+# Contributing
+
+This page lists the steps needed to set up a development environment and contribute to the project.
+
+1. Fork and clone this repo
```
gaborbernat: Please finish your sentences with punctuation.

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+
+2. Install `tox` via `pipx`: https://tox.wiki/en/4.11.3/installation.html#via-pipx
+
+3. Run tests
```
gaborbernat: `:`

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+   tox run
+   ```
+
+   or for a specific python version
+
+   ```shell
+   tox run -f py311
+   ```
+
+4. Running other tox commands (ex. linting)
```
gaborbernat: `:`

gaborbernat (changes requested):

src/filelock/_api.py
gaborbernat: Make it keyword only.

gaborbernat: Perhaps just `singleton` should suffice.

src/filelock/_api.py
```diff
@@ -109,6 +136,10 @@ def is_thread_local(self) -> bool:
         """:return: a flag indicating if this lock is thread local or not"""
         return self._is_thread_local
 
+    def is_singleton_per_lock_file(self) -> bool:
```
gaborbernat: is_singleton, make it a property

nefrob:
Downloading pypy locally to investigate test failures. ~~Not sure what the strategy for the type failures is (ex. switch from `Sequence` to `list`, drop the type entirely and noqa or resolve via getting type stubs for mypy?).~~ I misunderstoof mypy errors, it just wants type hints for the data inside the container objects.

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+   tox run
+   ```
+
+   or for a specific python version
+
+   ```shell
+   tox run -f py311
+   ```
+
+4. Running other tox commands (ex. linting):
```
gaborbernat: ex. -> eg.

src/filelock/_api.py
```diff
@@ -76,12 +77,41 @@ class ThreadLocalFileContext(FileLockContext, local):
...
 
-    def __init__(
+    _instances: ClassVar[WeakValueDictionary[str, BaseFileLock]] = WeakValueDictionary()
+
+    def __new__(  # noqa: PYI034
+        cls,
+        lock_file: str | os.PathLike[str],
+        is_singleton: bool = False,  # noqa: FBT001, FBT002
+        *args: Sequence[Any],  # noqa: ARG003
+        **kwargs: Mapping[str, Any],  # noqa: ARG003
```
gaborbernat: I don't think you need args/kwargs here 🤔

src/filelock/_api.py
```diff
@@ -76,12 +77,41 @@ class ThreadLocalFileContext(FileLockContext, local):
 class BaseFileLock(ABC, contextlib.ContextDecorator):
     """Abstract base class for a file lock object."""
 
-    def __init__(
+    _instances: ClassVar[WeakValueDictionary[str, BaseFileLock]] = WeakValueDictionary()
+
+    def __new__(  # noqa: PYI034
+        cls,
+        lock_file: str | os.PathLike[str],
+        is_singleton: bool = False,  # noqa: FBT001, FBT002
```
gaborbernat: do we need default here?

gaborbernat (changes requested):
Don't disable ruff checks, address them.

src/filelock/_api.py
nefrob: The signature of `__new__` should match that of `__init__` or handle the extra args/kwargs as is done here, otherwise instantiation will fail. But maybe I am missing something here...

Ref: https://docs.python.org/3/reference/datamodel.html#classes.

src/filelock/_api.py
nefrob: Yes, if I drop the default then the argument must be specified every time we do `FileLock(...)` unfortunately.

src/filelock/_api.py
```diff
@@ -76,12 +77,41 @@ class ThreadLocalFileContext(FileLockContext, local):
 class BaseFileLock(ABC, contextlib.ContextDecorator):
     """Abstract base class for a file lock object."""
 
-    def __init__(
+    _instances: ClassVar[WeakValueDictionary[str, BaseFileLock]] = WeakValueDictionary()
+
+    def __new__(  # noqa: PYI034
```
nefrob: On one more Ruff error ignored, if we change the return type to `Self` ruff is happy but mypy fails. I opted to satisfy mypy but ignore the ruff rule instead but perhaps there is a combined solution (return type `Self | BaseFileLock` works but seems undesirable)?

gaborbernat (approved):

nefrob:
Thanks so much for the help!

YouJiacheng:
I think there might be a bug: all derived class of BaseFileLock share the same `_instances`:
```python
from filelock import UnixFileLock, BaseFileLock

assert UnixFileLock._instances is BaseFileLock._instances
```
This can be resolved by meta class or simpler `__init_subclass__` (python>=3.6)

YouJiacheng:
BTW, maybe we should check if args like `mode` and `timeout` is the same as the singleton?

nefrob:
@YouJiacheng good callout. I think I avoided a metaclass in the first place because `BaseFileLock` inherits from `ABC` which sets its metaclass already. I can take a look at this later today.

Checking mode and timeout don't seem particularly relevant if we determine two singleton locks are the sam object (so will have the same values), unless you meant something else there.

YouJiacheng:
Okay, `__init_subclass__ ` can be used instead of metaclass.

By checking `mode` and `timeout`, I mean: if `mode` and `timeout` are different, but `lock_file` are the same, there should be some warning, since the second `mode` and `timeout` will be ignored.

nefrob:
@YouJiacheng here is the fix using `__init_subclass__`: https://github.com/tox-dev/filelock/pull/318.

As for the other question, deciding on how to handle that seems like something for @gaborbernat to decide on. We could:
1. raise an error
2. update the existing singleton lock
3. ignore doing anything (current behavior)

gaborbernat:
> 1. raise an error

Would be my vote.

nefrob:
Fix for incompatible singleton args: https://github.com/tox-dev/filelock/pull/320.

PR #318, opened 2024-03-25
Make singleton class instance dict unique per subclass

nefrob:
Re: https://github.com/tox-dev/filelock/pull/283#issuecomment-2016873026.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -8,7 +8,7 @@ import warnings
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from threading import local
-from typing import TYPE_CHECKING, Any, ClassVar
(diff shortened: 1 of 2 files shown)


src/filelock/_api.py
```diff
@@ -77,7 +77,7 @@ class ThreadLocalFileContext(FileLockContext, local):
 class BaseFileLock(ABC, contextlib.ContextDecorator):
     """Abstract base class for a file lock object."""
 
-    _instances: ClassVar[WeakValueDictionary[str, BaseFileLock]] = WeakValueDictionary()
```
gaborbernat, line 80: This is no longer a class variable, so we should change the annotation 👍

gaborbernat (changes requested):

gaborbernat (approved):

PR #320, opened 2024-03-26
Raise error on incompatible singleton timeout and mode args

nefrob:
Re: https://github.com/tox-dev/filelock/pull/283#issuecomment-2018116737.

This is a breaking change for singleton locks that subclass and implement their own context management.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -82,8 +82,8 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
     def __new__(  # noqa: PLR0913
         cls,
         lock_file: str | os.PathLike[str],
-        timeout: float = -1,  # noqa: ARG003
-        mode: int = 0o644,  # noqa: ARG003
+        timeout: float = -1,
+        mode: int = 0o644,
         thread_local: bool = True,  # noqa: ARG003, FBT001, FBT002
         *,
         is_singleton: bool = False,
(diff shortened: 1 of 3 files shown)


src/filelock/_api.py
```diff
@@ -169,6 +172,11 @@ def timeout(self, value: float | str) -> None:
         """
         self._context.timeout = float(value)
 
+    @property
+    def mode(self) -> int:
```
nefrob, line 181: note: we add this as a property to allow subclasses to override and implement their own context management.

tests/test_filelock.py
```diff
@@ -675,6 +675,17 @@ def test_singleton_locks_are_distinct_per_lock_file(lock_type: type[BaseFileLock
     assert lock_1 is not lock_2
 
 
+@pytest.mark.parametrize("lock_type", [FileLock, SoftFileLock])
+def test_singleton_locks_must_be_initialized_with_the__same_args(lock_type: type[BaseFileLock], tmp_path: Path) -> None:
+    lock_path = tmp_path / "a"
+    lock = lock_type(str(lock_path), is_singleton=True)  # noqa: F841
```
nefrob, line 682: note: noqa since we need the lock to exist for the scope of the test (as a singleton) even though it is "unused".

gaborbernat (approved):

PR #325, opened 2024-04-27
feat: `blocking` parameter on lock constructor with tests and docs

### LINKED ISSUE #64: try acquire filelock without blocking (timeout = 0)
Hi,

It seems that the latest docs (https://filelock.readthedocs.io/en/latest/) is not aligned with the code.
I wanted to try acquire a lock but without blocking if alreadly locked. for this to happen I expected the timeout param to ```acquire``` to be equal to 0 (as in many other locking primitives).
The docs says (acquire method of BaseFileLock) that 

> If timeout <= 0, there is no timeout and this method will block until the lock could be acquired

According to source code (both docstring of BaseFileLock.acquire and method source code) if timeout == 0 the method won't block.
BaseFileLock.acquire docstring:
```
:arg float timeout:
    The maximum time waited for the file lock.
    If ``timeout < 0``, there is no timeout and this method will
    block until the lock could be acquired.
    If ``timeout`` is None, the default :attr:`~timeout` is used.
```
BaseFileLock.acquire code:
```
elif timeout >= 0 and time.time() - start_time > timeout:
    logger().debug('Timeout on acquiring lock %s on %s', lock_id, lock_filename)
        raise Timeout(self._lock_file)
```

Why latest docs are not aligned - is it just not up-to-date? what needed in order to regenerate it to be aligned.

Thanks!

iamkhav:
Hey all, I'm Amon and this is my first PR here. Let me know if something's missing or not up to standards :)

### PR Summary
* Add `blocking` parameter on lock constructor. Similar in effect to `blocking` on `BaseFileLock.acquire`
* Add tests for a couple of different non-blocking lock cases. This also covers `timeout=0` which was the previous way to have a non-blocking lock when using context manager syntax.

I added some docs so this should close #64.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -156,7 +156,7 @@ You can pre-parametrize both of these options when constructing the lock for eas
             # do some work
             pass
     except Timeout:
-        print("Well, we tried once and couldn't acquire")
+        print("Well, we tried once and couldn't acquire.")
 
     lock_2 = FileLock("high_ground.txt.lock", timeout = 10)
     try:
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -113,7 +113,7 @@ The lock objects are recursive locks, which means that once acquired, they will 
 
 
 Timeouts and non-blocking locks
--------
+-------------------------------
 The :meth:`acquire <filelock.BaseFileLock.acquire>` method accepts a ``timeout`` parameter. If the lock cannot be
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):

iamkhav:
Hey @gaborbernat, thanks for the swift approval! I don't have rights to squash and merge this PR. Feel free to do so whenever it fits.

PR #332, opened 2024-06-03
asyncio support

Ovizro:
Add asyncio support for `filelock`
-------------

Usage:

```py
import asyncio
from filelock import AsyncFileLock

async def main() -> None:
    lock = AsyncFileLock("test.lock")
    async with lock:
        print("Lock acquired")
        await asyncio.sleep(1)

asyncio.run(main())
```

The default asynchronous approach in this implementation is to execute `_acquire` and `_release` in the thread pool.
Please ask if this is how you would like to see it implemented. If you think this is feasible, I will make further modifications and add tests for it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,10 +50,10 @@ optional-dependencies.testing = [
   "coverage>=7.3.2",
   "diff-cover>=8.0.1",
   "pytest>=7.4.3",
+  "pytest-asyncio",
   "pytest-cov>=4.1",
   "pytest-mock>=3.12",
   "pytest-timeout>=2.2",
-  "pytest-asyncio",
 ]
 optional-dependencies.typing = [
   "typing-extensions>=4.8; python_version<'3.11'",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ optional-dependencies.testing = [
   "coverage>=7.3.2",
   "diff-cover>=8.0.1",
   "pytest>=7.4.3",
-  "pytest-asyncio",
+  "pytest-asyncio>=0.21",
   "pytest-cov>=4.1",
   "pytest-mock>=3.12",
   "pytest-timeout>=2.2",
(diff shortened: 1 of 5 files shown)


gaborbernat (changes requested):
CI failing and no tests.

pyproject.toml
```diff
@@ -50,6 +50,7 @@ optional-dependencies.testing = [
   "coverage>=7.3.2",
   "diff-cover>=8.0.1",
   "pytest>=7.4.3",
+  "pytest-asyncio",
```
gaborbernat: Lower bound version like for everything else here.

src/filelock/asyncio.py
```diff
@@ -0,0 +1,327 @@
...
+
+        """
+        # Use the default timeout, if no timeout is provided.
+        if timeout is None:
+            timeout = self._context.timeout
+
+        if blocking is None:
+            blocking = self._context.blocking
+
+        if poll_intervall is not None:  # pragma: no cover
```
gaborbernat: Given this is a new class, let us remove this deprecated parts...

gaborbernat (changes requested):

tests/test_async_filelock.py
```diff
@@ -0,0 +1,181 @@
...
+        f"Attempting to release lock {id(lock)} on {lock_path}",
+        f"Lock {id(lock)} released on {lock_path}",
+    ]
+    assert [r.levelno for r in caplog.records] == [logging.DEBUG, logging.DEBUG, logging.DEBUG, logging.DEBUG]
+    assert [r.name for r in caplog.records] == ["filelock", "filelock", "filelock", "filelock"]
+    assert logging.getLogger("filelock").level == logging.NOTSET
+
+
+@pytest.mark.parametrize("lock_type", [AsyncFileLock, AsyncSoftFileLock])
+@pytest.mark.asyncio()
```
gaborbernat, line 76: Let us split this up so that one that is only testing one thing.

gaborbernat (changes requested):

tests/test_async_filelock.py
Ovizro: This doesn't seem to be easy to split. Is there a more detailed split plan? In fact, it is basically consistent with the unit test of synchronized file locks.

gaborbernat (approved):

PR #334, opened 2024-06-11
Don't initialize BaseFileLock when just returning existing instance

### LINKED ISSUE #331: is_singleton=True with nested lock hangs
I noticed the following code hangs. Is this not a supported usage?
```
from filelock import FileLock

with FileLock('.lock', is_singleton=True):
    with FileLock('.lock', is_singleton=True):
        ...

```

This works fine:
```
from filelock import FileLock

lock1 = FileLock('.lock', is_singleton=True)
lock2 = FileLock('.lock', is_singleton=True)
with lock1:
    with lock2:
        ...

```

Environment:
- Python 3.11.9
- filelock==3.14.0
- platform=MacOS Sonoma 14.5

ethanbb:
Fixes #331. Avoids re-initializing a `FileLock` or `BaseFileLock` if `is_singleton` is true and `__new__` just returns the singleton, which allows the lock to be re-acquired via the constructor if it has already been acquired without resetting the count.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -80,28 +80,33 @@ class ThreadLocalFileContext(FileLockContext, local):
 class BaseFileLock(ABC, contextlib.ContextDecorator):
     """Abstract base class for a file lock object."""
 
-    _instances: WeakValueDictionary[str, BaseFileLock]
+    _instances: WeakValueDictionary[str, Self]
 
     def __new__(  # noqa: PLR0913
         cls,
         lock_file: str | os.PathLike[str],
         timeout: float = -1,
         mode: int = 0o644,
-        thread_local: bool = True,  # noqa: ARG003, FBT001, FBT002
+        thread_local: bool = True,  # noqa: FBT001, FBT002
         *,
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

tamird:
This (via the 3.15.0 release) may be responsible for breaking virtualenv. See https://github.com/pypa/virtualenv/issues/2735.

tamird:
Ah this is API breaking (`__init__` no longer has the same signature).

ethanbb:
OK sorry, can fix - just need to put in a dummy `__init__` that takes `*args, **kwargs` and is a noop?

tamird:
Doesn't sound like an awesome idea to me because type safety is out the window. Sounds like a test that subclasses and overrides `__init__` would be prudent either way.

schmod:
~Can somebody yank this release from PyPi?  This is an ecosystem-breaking change (doubly so because few folks will have the installation of `virtualenv` itself behind a lockfile)~

Nevermind, I see that's already been done!  Many thanks!

PR #338, opened 2024-06-12
Hotfix: Restore __init__ method; more robust initialization for singleton locks

ethanbb:
Fixes issue with subclassing brought up in pypa/virtualenv#2735 (discussed in #334)

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -117,6 +117,9 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         super().__init_subclass__(**kwargs)
         cls._instances = WeakValueDictionary()
 
+    def __init__(self, *args, **kwargs):
+        pass  # for backwards compatibility (don't break super().__init__ calls)
+
     def _initialize(  # noqa: PLR0913
         self,
         lock_file: str | os.PathLike[str],
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -114,14 +114,7 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         super().__init_subclass__(**kwargs)
         cls._instances = WeakValueDictionary()
 
-<<<<<<< HEAD
     def __init__(  # noqa: PLR0913
-=======
-    def __init__(self, *args, **kwargs) -> None:
-        pass  # for backwards compatibility (don't break super().__init__ calls)
-
-    def _initialize(  # noqa: PLR0913
->>>>>>> 0d3d1cbd7830a436223d4e93712e7116237de86b
         self,
         lock_file: str | os.PathLike[str],
         timeout: float = -1,
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -114,22 +114,22 @@ class BaseFileLock(ABC, contextlib.ContextDecorator):
         super().__init_subclass__(**kwargs)
         cls._instances = WeakValueDictionary()
 
-<<<<<<< HEAD
+
+<< << << < HEAD
     def __init__(  # noqa: PLR0913
-=======
-    def __init__(self, *args, **kwargs) -> None:
+== == ===def __init__(self, *args, **kwargs) -> None:
         pass  # for backwards compatibility (don't break super().__init__ calls)
 
     def _initialize(  # noqa: PLR0913
->>>>>>> 0d3d1cbd7830a436223d4e93712e7116237de86b
+>> >> >> > 0d3d1cbd7830a436223d4e93712e7116237de86b
         self,
         lock_file: str | os.PathLike[str],
-        timeout: float = -1,
-        mode: int = 0o644,
-        thread_local: bool = True,  # noqa: FBT001, FBT002
+        timeout: float=-1,
+        mode: int=0o644,
+        thread_local: bool=True,  # noqa: FBT001, FBT002
         *,
-        blocking: bool = True,
-        is_singleton: bool = False,
+        blocking: bool=True,
+        is_singleton: bool=False,
     ) -> None:
         """
         Create a new lock object.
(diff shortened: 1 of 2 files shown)


src/filelock/_api.py
```diff
@@ -95,14 +95,11 @@ def __new__(  # noqa: PLR0913
...
-            self._initialize(lock_file, timeout, mode, thread_local, blocking=blocking, is_singleton=is_singleton)
-            return self
+            return super().__new__(cls)
 
         instance = cls._instances.get(str(lock_file))
         if not instance:
             self = super().__new__(cls)
-            self._initialize(lock_file, timeout, mode, thread_local, blocking=blocking, is_singleton=is_singleton)
             cls._instances[str(lock_file)] = self
             return self
```
tamird, line 105: the checks right below this line (can't anchor this comment there) should move to `__init__` and should be more general. currently `thread_local` and `blocking` aren't checked.

ethanbb:
Will add a test for this as well.

src/filelock/_api.py
```diff
@@ -143,6 +140,9 @@ def _initialize(  # noqa: PLR0913
             to pass the same object around.
 
         """
+        if hasattr(self, "_context"):
```
kwist-sgr: I wonder if this will work with [this](https://github.com/pypa/virtualenv/blob/429d6a2319c1bf917452ca418bf041fe0daa130d/src/virtualenv/util/lock.py#L16)?

ethanbb: Don't see why not, it's not setting the `_context` attribute is it? I also added a test for `is_singleton` along with some other tweaks in next push

ethanbb:
Sorry for this mess. I understand if you want to take longer to review this time; did not realize so much infrastructure relies on this package.

gaborbernat:
> Sorry for this mess. I understand if you want to take longer to review this time; did not realize so much infrastructure relies on this package.

No worries All good.

src/filelock/_api.py
```diff
@@ -143,6 +136,19 @@ def _initialize(  # noqa: PLR0913
...
+        if is_singleton and hasattr(self, "_context"):
+            # test whether other parameters match existing instance.
+            if (
+                self.is_thread_local() != thread_local
+                or not self.is_singleton
+                or self._context.timeout != timeout
+                or self._context.mode != mode
+                or self._context.blocking != blocking
+            ):
+                msg = "Singleton lock instances cannot be initialized with differing arguments"
```
tamird: this message is unfortunate - it doesn't tell the reader anything about the differences.

tamird (approved):
Looks reasonable.

src/filelock/_api.py
ethanbb: Yeah this can be expanded to give more details. Will update.

tamird: Complex enough to need a test now.

gaborbernat: Everything needs a test 😅 this project requires 100% code coverage.

ethanbb: Actually there is a test for it, which I will expand to include the other parameters, but it doesn't seem to mind that the error message has changed - guess that's not considered enough to fail a test? 

```python
with pytest.raises(ValueError, match="Singleton lock instances cannot be initialized with differing arguments"):
        lock_type(str(lock_path), timeout=10, is_singleton=True)
with pytest.raises(ValueError, match="Singleton lock instances cannot be initialized with differing arguments"):
        lock_type(str(lock_path), mode=0, is_singleton=True)
```

ethanbb: Oh never mind, I see that it doesn't have to be a full string match.

gaborbernat (changes requested):
Please add tests for the error message for each parameter differing 👍

src/filelock/_api.py
```diff
@@ -143,6 +136,30 @@ def _initialize(  # noqa: PLR0913
...
+                raise RuntimeError(msg)
+
+            params_to_check = {
+                "is_thread_local": (thread_local, self.is_thread_local()),
+                "timeout": (timeout, self.timeout),
+                "mode": (mode, self.mode),
+                "blocking": (blocking, self.blocking),
+            }
+
+            non_matching_params = {name: vals for name, vals in params_to_check.items() if vals[0] != vals[1]}
```
stasfilin: What if we will do something like 
```python
non_matching_params = {name: (passed_param, set_param) for name, (passed_param, set_param) in params_to_check.items() if passed_param != set_param}
if not non_matching_params:
        return
```

It will add a little bit more readable (better than `vals[0] != vals[1]`.

About `if not non_matching_params`:
1.  Using `not non_matching_params` is more straightforward and idiomatic.
2. `not non_matching_params` avoids an explicit function call to len(), making it marginally faster.

gaborbernat: I don't think we should worry hear about performance.

ethanbb: Sure, I agree with these changes.

ethanbb:
I had this test failure on `test_singleton_locks_are_deleted_when_no_external_references_exist` when testing locally as well and couldn't figure out what was wrong. Anyone have an idea?

gaborbernat (dismissed):

ethanbb:
> I had this test failure on `test_singleton_locks_are_deleted_when_no_external_references_exist` when testing locally as well and couldn't figure out what was wrong. Anyone have an idea?

Never mind, got it

gaborbernat (approved):

src/filelock/_api.py
kwist-sgr: This will cause the inherited class's `__init__` to be called again whenever `_CountedFileLock(is_singleton=True)` is called, which is at least not obvious (also, it'll lead to the re-creation of internal objects in the instance)

ethanbb: `_CountedFileLock` is not a singleton lock (always calls superclass constructor with the default is_singleton=False), so this is expected. Even if it were trying to create a singleton, I don't see how the superclass could force the subclass's `__init__` to be bypassed? That would also be a change in behavior from pre-bug (when even the superclass `__init__` wasn't bypassed).

kwist-sgr: > _CountedFileLock is not a singleton lock (always calls superclass constructor with the default is_singleton=False), so this is expected. 

Well, this's probably worth mentioning in the documentation if someone wants to make a singleton

> I don't see how the superclass could force the subclass's `__init__` to be bypassed?

It could be done with metaclass

ethanbb: > Well, this's probably worth mentioning in the documentation if someone wants to make a singleton

Agree that this parameter should be better documented, it tripped me up for a minute as well. 

> It could be done with metaclass

Ah interesting, that could be cleaner. I've never used metaclasses. As I understand, you would define new in the metaclass to incorporate any overridden init behavior when a new object is actually created, and make the real init empty. 

Maybe worth looking into @gaborbernat? but above my pay grade

PR #340, opened 2024-06-17
Use a metaclass to implement the singleton pattern

kwist-sgr:
## Reason

More correct implementation of the `singleton` pattern.

There is a problem with the current implementation: class's `__init__` to be called again whenever `lock_class(is_singleton=True)` is called, which is at least not obvious (also, it'll lead to the re-initializing instance's internal objects, e.g.: [virtualenv/util/lock.py](https://github.com/pypa/virtualenv/blob/429d6a2319c1bf917452ca418bf041fe0daa130d/src/virtualenv/util/lock.py#L16))

## What was done in this PR

- added classes `FileLockMeta`/`AsyncFileLockMeta` used as metaclasses for `BaseFileLock`/`BaseAsyncFileLock`
- added a test that checks that in the case of a singleton, the `__init__` method is called only once

### Related PRs

- tox-dev/filelock#338
- tox-dev/filelock#334

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -8,7 +8,7 @@ import warnings
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
 from threading import local
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 from weakref import WeakValueDictionary
 
 from ._error import Timeout
@@ -90,7 +90,7 @@ class FileLockMeta(ABCMeta):
         **kwargs: Any,  # capture remaining kwargs for subclasses  # noqa: ANN401
     ) -> BaseFileLock:
         if is_singleton:
-            instance = cls._instances.get(str(lock_file))
+            instance = cls._instances.get(str(lock_file))  # type: ignore[attr-defined]
             if instance:
                 params_to_check = {
                     "thread_local": (thread_local, instance.is_thread_local()),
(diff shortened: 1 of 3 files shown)


gaborbernat (approved):

gaborbernat (changes requested):
CI failing.

gaborbernat (approved):

gaborbernat:
@kwist-sgr https://github.com/tox-dev/sphinx-autodoc-typehints/actions/runs/9583644379/job/26425402570 seems broke the world had to yank it...

tamird:
To be fair it seems virtualenv is violating the contract by not implementing the proper signature of `__init__` [here](https://github.com/pypa/virtualenv/blob/bd91d48f01da390c8b3aa19b7b0dc04f34ec1e5d/src/virtualenv/util/lock.py#L16)

gaborbernat:
The thing is that all other args past the lockfile should be optional... or they were, so without making it a breaking change we can't suddenly require downstream to have this optional args...

ethanbb:
~OK now the issue is using the default `__new__` for `BaseFileLock`. Since `__init__` doesn't accept kwargs, `__new__` won't either. The way it previously worked was, if you wanted to add more arguments, you have to override `__init__` but not `__new__` since `__new__` explicitly allowed unused keyword arguments (maybe a weird pattern) but now that function is gone from `BaseFileLock`, so it's using the same signature as `__init__` by default.~

Hmm actually never mind, that doesn't make sense with the error

ethanbb:
OK I understand @gaborbernat's explanation. Yeah looking further, beyond just the super call, the `__call__` signature assumes that subclasses will all accept the same arguments (at minimum). And if that's not true, testing whether the singleton matches the passed-in parameters isn't possible at this point in the code in this way.

kwist-sgr:
@gaborbernat Oh. I apologize. I'm going to create a fix

PR #345, opened 2024-06-19
Fix `TypeError: _CountedFileLock.__init__() got an unexpected keyword argument 'timeout'`

kwist-sgr:
## Reason

Fixed an issue with the `BaseFileLock` subclass when the `__init__` method signature changes

### Issues

- tox-dev/filelock/issues/343

### Related PRs

- tox-dev/filelock#344
- tox-dev/filelock#340

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -116,7 +116,7 @@ class FileLockMeta(ABCMeta):
                 raise ValueError(msg)
 
         # Workaround to make `__init__`'s params optional in subclasses
-        # E.g. virtualenv changes the `__init__` signature of the `BaseFileLock` class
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

hmaarrfk:
Is this kind of behavior actually desirable?

It causes parameter to just "silently get dropped" and users may expect them to have an effect....

hmaarrfk:
I guess I think you should decide on the common signature you want to support. 

Does it include **kwargs?? 

Anyway, we hit this because we subclassed you guys and our signature was:


```
class Foo(FileLock):
     def __init__(self, *args, new_param=None, **kwargs):
         super().__init__(*args, **kwargs)
```

and the `timeout` parameter just got dropped.

In either case, we are just going to stop subclassing, but signature rewriting should really not be used for this kind of convenience stuff.

PR #351, opened 2024-07-21
Add logging for OSError exceptions

aamir-s18:
Hey, I had some issues with a library using filelock. My process was stucking all the time, it took me some time and printing messages in filelock to figure out that the issue was related to my mounted NFS storage. I think it would be helpful to have a log when `fcntl.flock` fails, so the user know where the issue is coming from. 

I added the two logging messages for Windows and Unix.

### DIFF
--- a/src/filelock/_unix.py
+++ b/src/filelock/_unix.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import os
 import sys
 from contextlib import suppress
@@ -9,6 +10,8 @@ from typing import cast
 
 from ._api import BaseFileLock
 from ._util import ensure_directory_exists
+
+_LOGGER = logging.getLogger("filelock")
 
 #: a flag to indicate if the fcntl API is available
 has_fcntl = False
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
test please

src/filelock/_unix.py
```diff
@@ -45,6 +48,7 @@ def _acquire(self) -> None:
             try:
                 fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
             except OSError as exception:
+                _LOGGER.debug("Failed to acquire lock: %s", exception)
```
asqui, line 51: It would be better to use `exc_info=True` to include full details of the exception (type, message, traceback) rather than just the message alone.

Like this:
```suggestion
                _LOGGER.debug("Failed to acquire lock", exc_info=True)
```

https://docs.python.org/3/library/logging.html#logging.Logger.debug

gaborbernat:
Seems stalled.

PR #398, opened 2025-02-19
Support fcntl check on Emscripten

juntyr:
In Pyodide/Emscripten, file locking is not (yet) supported (see e.g. https://github.com/pyodide/pyodide/discussions/4150#discussioncomment-7026664).

This PR extends the `fcntl` import check to also test whether the locking APIs are available, so that the soft locking fallback can be used if they are not. As a result, the `filelock` package becomes usable in Pyodide/Emscripten.

### DIFF
--- a/src/filelock/_unix.py
+++ b/src/filelock/_unix.py
@@ -26,6 +26,7 @@ if sys.platform == "win32":  # pragma: win32 cover
 else:  # pragma: win32 no cover
     try:
         import fcntl
+
         _ = (fcntl.lock, fcntl.LOCK_EX, fcntl.LOCK_NB, fcntl.LOCK_UN)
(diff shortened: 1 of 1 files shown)


gaborbernat (changes requested):
Seems the CI does not like your changes 🤔

juntyr:
> Seems the CI does not like your changes 🤔

Whoops, I managed to mistype `fcntl.flock`, I hope we'll get a green CI now :)

gaborbernat (approved):

juntyr:
@gaborbernat Is there a rough timeline for when the next version of filelock will be released?

gaborbernat:
https://github.com/tox-dev/filelock/releases/tag/3.18.0

PR #399, opened 2025-02-19
✨ feat(lock): add SQLite-based ReadWriteLock

leventov:
The existing lock classes (`FileLock`, `SoftFileLock`) only support exclusive locking -- every caller blocks every other caller, even when most only need to read. This is fine for write-heavy workloads, but read-heavy ones pay an unnecessary serialization cost. 🔒

This PR adds `ReadWriteLock`, a cross-process shared/exclusive lock backed by SQLite transactions. Multiple processes can hold a read lock simultaneously while write locks remain exclusive. The lock is reentrant within the same mode (nested reads or nested writes from the same thread), but upgrading from read to write or downgrading from write to read raises `RuntimeError` to prevent deadlocks. Write locks are pinned to the acquiring thread.

The implementation uses SQLite's built-in `BEGIN TRANSACTION` (shared) and `BEGIN EXCLUSIVE TRANSACTION` (exclusive) with `journal_mode=MEMORY` rather than WAL, because WAL would allow readers to proceed during an exclusive write -- exactly the behavior we need to prevent. A double-check locking pattern serializes concurrent threads through a `threading.Lock` before touching SQLite, and `busy_timeout` is recomputed after each potentially-blocking pragma to keep timeout accounting accurate. Instances are singletons by default (keyed by resolved path via a `WeakValueDictionary`), matching the pattern established by `BaseFileLock`'s `is_singleton` parameter.

📝 The docs (`docs/index.rst`) have been substantially reworked: a new section hierarchy (`Usage` / `Lock types`), platform resolution details for `FileLock`, a dedicated `ReadWriteLock` section with usage examples, limitations for every lock type, async variant mentions in "Choosing the right lock", and singleton lock documentation. The changelog has been backfilled from v3.12.1 through v3.20.3. Code examples now use `pathlib.Path` and descriptive variable names throughout. Existing docstrings in `_api.py`, `asyncio.py`, and `_soft.py` have been expanded with `:param:` / `:raises:` / `:return:` tags and `versionadded` directives.

♻️ The `_read_write.py` source extracts shared acquire logic into `_acquire()`, `_validate_reentrant()`, and `_configure_and_begin()` to eliminate ~50 lines of duplication between `acquire_read` and `acquire_write`. Internal attributes use underscore-prefixed names (`_con`, `_lock_level`, etc.) and imports are relative throughout.

A PyPy-specific `test_virtualenv` skip was added due to a segfault in PyPy's `_sqlite3.Cursor.__del__._reset` during garbage collection -- no upstream issue was found for this bug.


src/filelock/_read_write.py
```diff
@@ -0,0 +1,80 @@
...
+if TYPE_CHECKING:
+    import os
+
+
+class _SQLiteLock(BaseFileLock):
+    def __init__(self, lock_file: str | os.PathLike[str], timeout: float = -1, blocking: bool = True) -> None:
+        super().__init__(lock_file, timeout, blocking)
+        self.procLock = threading.Lock()
+        self.con = sqlite3.connect(self._context.lock_file, check_same_thread=False)
+        # Redundant unless there are "rogue" processes that open the db
```
gaborbernat: Please align comments to 120 charachters 🙇

src/filelock/_read_write.py
```diff
@@ -0,0 +1,80 @@
...
+from _error import Timeout
+
+from filelock._api import BaseFileLock
+
+if TYPE_CHECKING:
+    import os
+
+
+class _SQLiteLock(BaseFileLock):
+    def __init__(self, lock_file: str | os.PathLike[str], timeout: float = -1, blocking: bool = True) -> None:
```
gaborbernat: Prefer 
```
from os import PathLike
```

leventov:
@gaborbernat I've pushed a more "serious" attempt.

gaborbernat (changes requested):
Looks better, can you add tests please?

leventov:
@gaborbernat what do you think about using [SQLite's VFS interface](https://www.sqlite.org/vfs.html) directly, without the "dummy database" thing? https://github.com/rogerbinns/apsw provides bindings

gaborbernat:
It's important that we have no dependencies. The only alternative could be
to hide this functionality behind some extra, I do not know how widely
available the bindings are in form of wheels.

leventov:
@gaborbernat I've added tests.

If you want to finesse API (context managers vs. parameters to acquire_read() and acquire_write() methods; are read_lock() and write_lock() methods even needed? Are instance's default timeout and blocking needed?), docs, formatting, class/method naming, etc., could you please do it yourself to your liking?

gaborbernat (changes requested):
CI failing.

leventov:
@gaborbernat I don't need this PR to be merged to use it later through `filelock` library. In fact, I submitted a PR only to attract code review and discussion. Myself, I'm going to use this code copied (and modified to an extent) in the codebase. So, I'm not invested in the merge per se.

What of these seem valid to you, and what seem like false positives?
```
src/filelock/_read_write.py:61: error: Returning Any from function declared to return "ReadWriteLock"  [no-any-return]
src/filelock/_read_write.py:61: error: "_ReadWriteLockMeta" has no attribute "get_lock"  [attr-defined]
src/filelock/_read_write.py:62: error: Returning Any from function declared to return "ReadWriteLock"  [no-any-return]
src/filelock/_read_write.py:67: error: Need type annotation for "_instances"  [var-annotated]
src/filelock/_read_write.py:77: error: Argument 2 for "super" not an instance of argument 1  [misc]
src/filelock/_read_write.py:91: error: Returning Any from function declared to return "ReadWriteLock"  [no-any-return]
```

Can the [CI auto-fix](https://github.com/tox-dev/filelock/pull/399/commits/0f8302d1311884a9f9ab4782470a63483630d21f) from `"ReadWriteLock"` to `ReadWriteLock` itself cause some of these errors?

[check / test pypy3.10 - windows-latest](https://github.com/tox-dev/filelock/actions/runs/13654464624/job/38575451882?pr=399) is test-suite related, perhaps even CI environment-related.

gaborbernat:
Moved to draft until we fix the issues.

PR #317, opened 2024-03-25
[BugFix] fix permission denied error when lock file is placed in `/tmp`

kota-iizuka:
When you put the lock file in a directory with sticky bit (like `/tmp`), you will get a permission denied error if you use a lock file created by someone else.

Work around this issue by excluding `os.O_CREAT` from `open_flags` if the lock file already exists.

cf. https://github.com/vllm-project/vllm/pull/3599#issuecomment-2017125950

#### sample code

```py
# sample.py
import filelock
with filelock.FileLock("tmp/a.lock", mode=0o666):
    pass
```

```sh
# create tmp file with sticky bit
$ mkdir tmp
$ sudo chmod 1777 tmp
$ sudo chown root:root tmp

# add lock file by another user
$ touch tmp/a.lock
$ sudo chmod 666 tmp/a.lock 
$ sudo chown 1111:1111 tmp/a.lock

# run sample code
$ python sample.py 
Traceback (most recent call last):
  File "/workspaces/filelock/sample.py", line 4, in <module>
    with filelock.FileLock(lock_path, mode=0o666):
  File "/workspaces/filelock/src/filelock/_api.py", line 301, in __enter__
    self.acquire()
  File "/workspaces/filelock/src/filelock/_api.py", line 257, in acquire
    self._acquire()
  File "/workspaces/filelock/src/filelock/_unix.py", line 41, in _acquire
    fd = os.open(self.lock_file, open_flags, self._context.mode)
PermissionError: [Errno 13] Permission denied: 'tmp/a.lock'
```

PR #408, opened 2025-04-02
Enable removal of UNIX lock files

### LINKED ISSUE #31: Random assertions failing in test.py
I'm maintaining py-filelock for Debian (https://packages.qa.debian.org/p/python-filelock.html). We are running`test.py` in a clean chroot during the package build to verify that all tests pass in the version to be packaged. However, we recently noticed that current versions of the library fail their `test.py` runs with random assertions being hit:

```
PYTHONPATH=. python2 /build/python-filelock-3.0.0/debian/test.py
.........FF..F........
======================================================================
FAIL: test_threaded1 (__main__.FileLockTest)
----------------------------------------------------------------------
  File "/build/python-filelock-3.0.0/debian/test.py", line 233, in test_threaded1
    threads1[i].join()
  File "/build/python-filelock-3.0.0/debian/test.py", line 66, in join
    raise (wrapper_ex.__class__, wrapper_ex, self.ex[2])
AssertionError

======================================================================
FAIL: test_timeout (__main__.FileLockTest)
----------------------------------------------------------------------
  File "/build/python-filelock-3.0.0/debian/test.py", line 253, in test_timeout
    self.assertRaises(filelock.Timeout, lock2.acquire, timeout=1)
AssertionError: Timeout not raised

======================================================================
FAIL: test_default_timeout (__main__.SoftFileLockTest)
----------------------------------------------------------------------
  File "/build/python-filelock-3.0.0/debian/test.py", line 278, in test_default_timeout
    self.assertRaises(filelock.Timeout, lock2.acquire)
AssertionError: Timeout not raised

----------------------------------------------------------------------
Ran 22 tests in 6.045s
```

Patching the source to revert cb1d83d6cb18cc60fb6249fa59e2b47631161244 seems to reliably fix the issue but I'm not sure stray lock files are left lying around. Any comments?

sbc100:
This avoids leaving orphan locks files around after we are done when them.

See https://github.com/emscripten-core/emscripten/pull/24035

Fixes: #31

### DIFF
--- a/src/filelock/_unix.py
+++ b/src/filelock/_unix.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import os
 import sys
 from contextlib import suppress
-from errno import ENOSYS
+from errno import EAGAIN, ENOSYS, EWOULDBLOCK
 from pathlib import Path
 from typing import cast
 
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
Please add tests 👍

sbc100:
> Please add tests 👍

I can add a test that confirm the that lock file is deleted.

Added a test the races mentioned do not pose a problem is harder.   Even if we added some intense stress tests that right conditions for the race might not happen.

Hopefully the former is enough?

gaborbernat (dismissed):

src/filelock/_unix.py
```diff
@@ -38,9 +38,7 @@ class UnixFileLock(BaseFileLock):
 
         def _acquire(self) -> None:
             ensure_directory_exists(self.lock_file)
-            open_flags = os.O_RDWR | os.O_TRUNC
-            if not Path(self.lock_file).exists():
-                open_flags |= os.O_CREAT
+            open_flags = os.O_RDWR | os.O_TRUNC | os.O_CREAT
```
sbc100: I worry a little that I had to remove this workaround which was added in: #317.

I can't seem to find a way to keep both of these fixes in place.

gaborbernat: Yeah, I'm hesitant to approve it without it though 🤔

saikishor:
@gaborbernat When can we expect this to be merged and released?

gaborbernat:
> I worry a little that I had to remove this workaround which was added in: #317.
> 
> I can't seem to find a way to keep both of these fixes in place.

Not until we address this.

PR #438, opened 2025-08-22
Update docs with example

### LINKED ISSUE #402: Confusing but perhaps deliberate behaviour ?
Hi there, thanks for the package. A perhaps naive question, but the following behaviour seemed unintuitive to me. In short, if you chain method calls and don't assign the result to a variable, you don't acquire the lock. However, if you do assign the result, even to `_`, you acquire the lock. I must be missing something, but I can't tell what.

A minimum reproducible example is below.

```python
import tempfile
from pathlib import Path

import filelock

lock_file = Path(tempfile.mkdtemp()) / "index.json.lock"

# For some reason, this doesn't cause the lock to actually be acquired.
filelock.SoftFileLock(lock_file).acquire()
print(f"{lock_file.exists()=}")

# But this does
_ = filelock.SoftFileLock(lock_file).acquire()
print(f"{lock_file.exists()=}")
```

For what it's worth, I suspect the same underlying confusion causes this and #400. (#400 seems to also be a confusion about the flow, because the `acquire` method does take `polling_interval` as an argument, but it is true that `__enter__` does not. Maybe the answer is, "Always use `acquire` as a context manager, never antying else". If that is the answer, all fine, but it might be worth clarifying and updating the docs as current examples suggeset that using `acquire` not as a context manager is an expected use case too.)

znichollscr:
Closes #402

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -114,7 +114,10 @@ The lock objects are recursive locks, which means that once acquired, they will 
 It should also be noted that the lock is released during garbage collection.
 Put another way, if you do not assign an acquired lock to a variable,
 the lock will eventually be released (implicitly, not explicitly).
-This is explained with code below:
+For this reason, using the lock in the way shown below is not something you should ever do,
+always use the context manager (`with` form) instead.
+
+This issue is illustrated with code below:
 
 .. code-block:: python
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -111,6 +111,31 @@ The lock objects are recursive locks, which means that once acquired, they will 
         cite2()
     # And released here.
 
+It should also be noted that the lock is released during garbage collection.
+Put another way, if you do not assign an acquired lock to a variable,
+the lock will eventually be released (implicitly, not explicitly).
+This is explained with code below:
+
+.. code-block:: python
+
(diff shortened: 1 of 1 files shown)


znichollscr:
@gaborbernat how is this? Also please let me know if I need to do changelog etc. (if the answer is, 'rtfd', then just tell me that)

docs/index.rst
```diff
@@ -111,6 +111,31 @@ The lock objects are recursive locks, which means that once acquired, they will
...
+    import tempfile
+    from pathlib import Path
+
+    import filelock
+
+    # If you create a lock and acquire it but don't assign it,
+    # you will not actually hold the lock forever.
+    # Instead, the lock is released
+    # when the created variable is garbage collected.
+    FileLock(lock_path).acquire()
```
gaborbernat, line 133: Can you add a disclaimer that therefore no one ever should use it like this, prefer using the `with` form.

gaborbernat (changes requested):

docs/index.rst
znichollscr: Done (although on a different line, hope that is ok)

gaborbernat (approved):

PR #463, opened 2025-12-18
Support Unix systems without O_NOFOLLOW

mwilliamson:
Acquiring a lock uses the `O_NOFOLLOW` flag when calling `open`. However, this flag is not available on all platforms, such as GraalPy, causing an `AttributeError` to be raised when attempting to acquire a lock. According to the Python docs:

> The above constants are extensions and not present if they are not defined by the C library.

This handles such platforms by checking for the presence of the `O_NOFOLLOW` flag first.


gaborbernat (approved):
LGTM

mwilliamson:
Thanks!

PR #473, opened 2026-02-12
🐛 fix: gracefully handle missing `sqlite3` when importing `ReadWriteLock`

bayandin:
`filelock 3.21.0` added `ReadWriteLock` backed by SQLite and imported it unconditionally in `__init__.py`. This broke `import filelock` on Python installations where `_sqlite3` was not compiled (e.g. pyenv-built Python without `libsqlite3-dev`), even when `ReadWriteLock` was never used.

Wrap the import in a try/except so that `ReadWriteLock` is set to `None` when `sqlite3` is unavailable, allowing all other lock types to work as before.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -236,7 +236,8 @@ This library provides several lock implementations. Pick the one that matches yo
 
   *Limitations*: higher overhead than ``FileLock`` due to SQLite transactions. Creates a ``.db`` file on disk.
   No async variant. Cannot upgrade a read lock to a write lock (or vice versa). May not work on network filesystems
-  where SQLite locking is unreliable.
+  where SQLite locking is unreliable. Requires the ``sqlite3`` standard library module (unavailable if Python was built
+  without SQLite support).
 
 .. warning::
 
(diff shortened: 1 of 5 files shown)


gaborbernat (dismissed):

eschiff:
I don't think this works yet, we'd need to add an `except ImportError` for this to work properly, right?

gaborbernat:
> I don't think this works yet, we'd need to add an `except ImportError` for this to work properly, right?

This should be correct, what do you get? What's your stack?

eschiff:
docker run --rm --platform linux/amd64 ubuntu-noble-20250127-python3.11.11:0.0.7 sh -c "pip install filelock==3.21.1 >/dev/null 2>&1 && python3 -c 'import filelock; print(\"filelock 3.21.1: OK\")' 2>&1"

  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.11/site-packages/filelock/__init__.py", line 19, in <module>
    from ._read_write import ReadWriteLock
  File "/usr/local/lib/python3.11/site-packages/filelock/_read_write.py", line 7, in <module>
    import sqlite3
  File "/usr/local/lib/python3.11/sqlite3/__init__.py", line 57, in <module>
    from sqlite3.dbapi2 import *
  File "/usr/local/lib/python3.11/sqlite3/dbapi2.py", line 27, in <module>
    from _sqlite3 import *
ImportError: libsqlite3.so.0: cannot open shared object file: No such file or directory

PR #497, opened 2026-02-20
Fix ValueError in _acquire_transaction_lock when blocking=False with timeout

bysiber:
`ReadWriteLock._acquire_transaction_lock()` crashes with `ValueError` when called with `blocking=False` and a non-default timeout.

Python's `threading.Lock.acquire()` raises `ValueError: can't specify a timeout for a non-blocking call` when both `blocking=False` and a timeout are provided. The current code only special-cases `timeout == -1`, so any other timeout value combined with `blocking=False` hits this error:

```python
from filelock import ReadWriteLock

lock = ReadWriteLock("test.db")
lock.acquire_read(timeout=5, blocking=False)
# ValueError: can't specify a timeout for a non-blocking call
```

The fix handles `blocking=False` as a separate branch that never passes a timeout argument, since a timeout is meaningless for non-blocking acquisition anyway.

### DIFF
--- a/src/filelock/_read_write.py
+++ b/src/filelock/_read_write.py
@@ -159,11 +159,12 @@ class ReadWriteLock(metaclass=_ReadWriteLockMeta):
             _all_connections.add(self._con)
 
     def _acquire_transaction_lock(self, *, blocking: bool, timeout: float) -> None:
-        if timeout == -1:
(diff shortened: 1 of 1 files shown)


gaborbernat:
Thanks for the fix\! I've added a test and fixed lint issues, and pushed directly to https://github.com/tox-dev/filelock/tree/fix/transaction-lock-nonblocking-timeout

PR #534, opened 2026-04-19
✨ feat(soft): enable stale lock detection on Windows

gaborbernat:
`SoftFileLock` disabled stale lock detection on Windows because Windows recycles PIDs aggressively. A crashed process could leave a lock file behind, and a new unrelated process inheriting the same PID would fool `OpenProcess` into thinking the lock was still held. This caused permanent deadlocks after crashes on Windows (#531).

Separately, malformed lock files (empty, partial, corrupted) were silently ignored by `_try_break_stale_lock`, also causing permanent deadlocks on all platforms (#532). `SoftReadWriteLock` already handled this correctly by evicting unparseable markers.

🔐 On Windows, the lock file now stores the process creation FILETIME (via `kernel32.GetProcessTimes`) as a third line. On contention, if the PID is alive but its creation time differs from the stored value, the lock is recognized as stale from a recycled PID and evicted. Both the old 2-line format and the new 3-line format are accepted when reading, so mixed-version deployments degrade gracefully (old processes stay conservative and never evict new-format files).

Malformed lock files are now evicted when their `mtime` exceeds 2 seconds. ✨ The gap between `os.open` (file creation) and `os.write` (content write) is nanoseconds, so any malformed file older than 2 seconds is definitively from a crashed process. Fresh malformed files are preserved to avoid racing a holder whose `_write_lock_info` silently failed.

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -113,8 +113,8 @@ This is the same concept as a traditional **PID lock file** (as used by Unix dae
 library's ``PIDLockFile``). The PID stored in the file enables two important capabilities: identifying the lock holder
 via the :attr:`~filelock.SoftFileLock.pid` property, and detecting stale locks when the holding process has died.
 
-On Unix/macOS, processes can check if the lock holder is still alive and break stale locks automatically. On Windows,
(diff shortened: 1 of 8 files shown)


brentDSL:
Thank you for the quick response and fix!

PR #542, opened 2026-05-20
chore: improve filelock maintenance path

lphuc2250gma:
Summary:
- Tighten typing and add an edge-case unit test for Timeout exception: ensure Timeout.__reduce__ round-trips via pickle and that str(Timeout(lock_file)) format is stable; add explicit return type annotations where missing in _error.py.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/src/filelock/_error.py
+++ b/src/filelock/_error.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+
 class Timeout(TimeoutError):  # noqa: N818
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):

gaborbernat (approved):
LGTM

PR #545, opened 2026-05-24
chore: improve filelock maintenance path

lphuc2250gma:
Summary:
- Add unit tests covering the Timeout exception's __reduce__/pickling behavior and __repr__ formatting edge cases (e.g., empty lock_file path, non-ASCII paths) to lock in current behavior of the Timeout class in _error.py
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/src/filelock/_error.py
+++ b/src/filelock/_error.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):
LGTM

PR #548, opened 2026-06-03
🐛 fix(soft): refuse to follow symlinks when reading the lock file

dxbjavid:
SoftFileLock reads the lock file with Path.read_text in both the pid getter and _try_break_stale_lock, which follows symlinks and reads unbounded content. Once a holder dies, anyone able to write in the lock directory can replace the stale lock file with a symlink, so a peer running stale detection opens the link target (a FIFO that stalls the read, or a huge file). The new _read_lock_file helper opens with O_NOFOLLOW and caps the read, matching _read_marker in the soft read-write lock.


gaborbernat (approved):

dxbjavid:

PR #549, opened 2026-06-04
open marker reads non-blocking to refuse attacker-placed fifo

dxbjavid:
_read_marker in the soft read/write lock opens marker files with O_RDONLY | O_NOFOLLOW but no O_NONBLOCK, which is the same gap #548 closed in _soft._read_lock_file. I was going through the soft_rw markers after that change and noticed O_NOFOLLOW only rejects a symlink, not a real FIFO, so a non-cooperating same-UID process can pre-create a named pipe at a predictable marker path like foo.lock.write and any peer that reads it during stale-break or slot-claim blocks forever on the open, all while holding the state SoftFileLock. Adding O_NONBLOCK makes the open return at once so a hostile FIFO reads as a malformed marker and gets evicted like any other stale entry.


gaborbernat (approved):

dxbjavid:

PR #550, opened 2026-06-05
use lstat when checking lifetime so a symlinked lock file is not followed

dxbjavid:
reading the code I noticed _try_break_expired_lock uses Path(self.lock_file).stat() to decide if a lock is expired, which follows symlinks. a same-host process with write access to the lock directory can swap a held lock file for a symlink onto an old file, so stat sees the stale target mtime and a waiter breaks the live lock and acquires it, so two holders coexist. lstat reads the symlink's own mtime instead, matching the O_NOFOLLOW lock-file reads already in _soft.py and _soft_rw.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -368,7 +368,11 @@ class BaseFileLock(contextlib.ContextDecorator, metaclass=FileLockMeta):
         if (lifetime := self._context.lifetime) is None:
             return
         with contextlib.suppress(OSError):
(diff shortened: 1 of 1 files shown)


gaborbernat:
Superseded by #551, which keeps your `lstat` fix and adds the related hardening (verified-inode breaking, self-healing of malformed lock files, async executor cleanup) found while reviewing this. Thanks for opening this and catching the original symlink issue, @dxbjavid! 🙏

PR #553, opened 2026-06-08
check hostname in is_lock_held_by_us

dxbjavid:
is_lock_held_by_us compares only the pid from the lock file against os.getpid() and ignores the hostname stored alongside it, so on a shared filesystem where the same pid is live on two hosts a lock genuinely held by another node is reported as held by us, and a caller acting on that by releasing or breaking it can drop a lock another host still holds. the lock file already records the hostname for exactly this disambiguation and the stale-break path checks it, so this reads the holder through _parse_lock_holder and requires both the pid and the hostname to match before returning true. added a regression case to the existing is_lock_held_by_us parametrisation covering a same-pid, different-host file.

### DIFF
--- a/src/filelock/_soft.py
+++ b/src/filelock/_soft.py
@@ -159,10 +159,15 @@ class SoftFileLock(BaseFileLock):
         """
         Whether this lock is held by the current process.
 
-        :returns: ``True`` if the lock file exists and contains the current process's PID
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):
LGTM

dxbjavid:

PR #556, opened 2026-06-09
validate pid range in _parse_lock_holder

dxbjavid:
a soft lock file lives in a shared directory that a same-uid or cross-host peer can write, and _parse_lock_holder turns its first line into a pid with int() but never checks the range, unlike _parse_marker_bytes which already rejects anything outside 1..2**31-1. a holder line of 0 or -1 parses fine, so _is_process_alive ends up calling os.kill(0, 0) or os.kill(-1, 0), and those probe the caller's own process group rather than a real holder and report success, so a long-dead lock reads as still held and is never reclaimed. a pid too large for the C pid_t is worse: os.kill raises OverflowError, which is neither OSError nor ValueError, so it slips past the suppress() guard and aborts acquire() instead of letting the malformed lock self-heal.

the fix range-checks the parsed pid against the same window the marker parser uses, so an out-of-range value is treated as an unparseable holder and self-heals after the age threshold like any other malformed file. added a regression covering 0, -1 and an oversized pid.

### DIFF
--- a/src/filelock/_soft.py
+++ b/src/filelock/_soft.py
@@ -149,9 +149,9 @@ class SoftFileLock(BaseFileLock):
 
         """
         with suppress(OSError, ValueError):
-            content, _ = _read_lock_file(self.lock_file)
-            if content and (lines := content.strip().splitlines()):
-                return int(lines[0])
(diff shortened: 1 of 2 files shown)


dxbjavid:
Thanks for the review and merge. Appreciated.

PR #561, opened 2026-06-11
verify inode in break_lock_file before unlinking a stale lock

dxbjavid:
break_lock_file renames a lock it judged stale aside and then re-checks only the modification time before unlinking it, but on a filesystem with coarse mtime granularity like nfs or fat a peer can break and recreate the lock within the same second, so the live recreation carries the same mtime the waiter saw and gets unlinked, letting a second process acquire it. the sibling marker break in _soft_rw already re-verifies the marker token for exactly this reason, so this captures the inode at the stale decision and aborts the break when the renamed file's inode no longer matches. i added a regression test that recreates the lock with a fresh inode under an identical mtime, which unlinks the live file before this change and leaves it untouched after.


gaborbernat (approved):
LGTM

dxbjavid:

PR #562, opened 2026-06-13
keep the read/write heartbeat alive on a transient touch error

dxbjavid:
SoftReadWriteLock._refresh_marker only catches FileNotFoundError around its heartbeat touch, so a transient ESTALE or EIO (routine on the NFS-style filesystems this lock is meant for) escapes as an unhandled OSError and kills the heartbeat thread without ever setting heartbeat_stop. The holder still believes it holds the lock while the marker stops being refreshed, so it goes stale, a peer evicts it and acquires, and two processes end up holding conflicting locks. The writer-drain touch in readers_drained_touching already wraps the same call in suppress(OSError), so this just brings the heartbeat into line: a transient touch error is swallowed and retried on the next tick, since the read just above has already confirmed the marker is still ours and a genuinely vanished marker is caught by the next iteration's _read_marker.


gaborbernat (approved):

dxbjavid:

PR #563, opened 2026-06-14
serialise read/write release rollback against a concurrent acquire

dxbjavid:
ReadWriteLock.release() rolls back the transaction on the shared sqlite3 connection outside _transaction_lock, even though acquire() runs its own BEGIN under that lock. Once release has put lock_level back to 0 inside the internal lock but has not yet rolled back, another thread acquiring on the same instance sees the level at 0, takes _transaction_lock and issues BEGIN against the connection whose transaction is still open, which raises "cannot start a transaction within a transaction" (re-raised, since it is not a "database is locked" timeout); in the reverse ordering the rollback instead ends the new holder's freshly started transaction and quietly drops the database lock while it still thinks it holds it. The added test parks the rollback with the level already at 0 and shows the racing acquirer now waits rather than erroring. Wrapping the rollback in _transaction_lock keeps it serialised against acquire the same way the rest of the transaction work already is.

### DIFF
--- a/tests/test_read_write_unit.py
+++ b/tests/test_read_write_unit.py
@@ -577,6 +577,53 @@ def test_cleanup_connections_closes_all(tmp_path: Path) -> None:
         con2.execute("SELECT 1;")
 
 
+def test_release_rollback_serialised_against_concurrent_acquire(lock_file: str) -> None:
+    # release() rolls back the transaction on the shared connection; a concurrent acquire on another thread
+    # must not be able to BEGIN while that rollback is still in flight. Park the rollback with the lock level
+    # already back at 0 and confirm the racing acquire waits for it instead of erroring with "cannot start a
+    # transaction within a transaction".
+    lock = ReadWriteLock(lock_file, is_singleton=False)
+    lock.acquire_read()
+
+    started = threading.Event()
+    proceed = threading.Event()
+    real_con = lock._con
+
+    class _SlowRollbackCon:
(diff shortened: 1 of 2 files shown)


gaborbernat (changes requested):
CI failing

dxbjavid:
sorry about that. the failures were all in the new code: codespell didn't like the en-GB spelling in my comment, ty wanted its own ignore syntax rather than the mypy one i'd used, and ruff flagged the helper rollback as no-self-use. fixed all three and the unit suite is green again locally.

tests/test_read_write_unit.py
```diff
@@ -577,6 +577,53 @@ def test_cleanup_connections_closes_all(tmp_path: Path) -> None:
...
+    # already back at 0 and confirm the racing acquire waits for it instead of erroring with "cannot start a
+    # transaction within a transaction".
+    lock = ReadWriteLock(lock_file, is_singleton=False)
+    lock.acquire_read()
+
+    started = threading.Event()
+    proceed = threading.Event()
+    real_con = lock._con
+
+    class _SlowRollbackCon:
```
gaborbernat, line 592: Don't ignore either the linter or type checker. Fix the errors.

gaborbernat (changes requested):

gaborbernat (approved):
LGTM

dxbjavid:

PR #565, opened 2026-06-16
🔒 fix(soft_rw): refresh held marker through the verified fd

dxbjavid:
The heartbeat that keeps a soft read/write lock alive refreshed its marker by re-opening the path, so a same-uid peer could replace the marker in the window between the verified read and the refresh and redirect the timestamp update onto a file the lock never owned. 🔒 Every other marker access in this module opens with `O_NOFOLLOW` to refuse a swapped symlink, and the refresh was the one place that trusted the path a second time.

The refresh now opens the marker once with `O_NOFOLLOW`, confirms the stored token on that descriptor, and runs `os.utime` against the same descriptor. Because the touch targets the open inode instead of re-resolving the name, the swap window closes for a regular-file replacement as well as a symlink, going beyond the `follow_symlinks=False` guard that covered only the symlink case. A failed open now counts as eviction and stops the heartbeat rather than retrying.

Where `os.utime` cannot operate on a descriptor the refresh falls back to the path-based call that still refuses to follow a symlink, so behavior on those platforms is unchanged. There are no API or configuration changes.

### DIFF
--- a/src/filelock/_soft_rw/_sync.py
+++ b/src/filelock/_soft_rw/_sync.py
@@ -33,6 +33,9 @@ _BREAK_SUFFIX = ".break"
 _MAX_MARKER_SIZE = 1024
 _O_NOFOLLOW = getattr(os, "O_NOFOLLOW", 0)
 _O_NONBLOCK = getattr(os, "O_NONBLOCK", 0)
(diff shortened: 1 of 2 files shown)


dxbjavid:

PR #566, opened 2026-06-18
only unlink the writer marker on release if it is still ours

dxbjavid:
If a writer is paused past stale_threshold (a long GC pause, SIGSTOP, a suspended VM) a peer evicts its now-stale .write marker, claims the writer slot and writes its own marker at the same path; the original holder still has _hold set, so on release it unlinks .write by path and deletes the peer's live marker, letting a second writer in and breaking mutual exclusion. The phase-2 timeout cleanup in _acquire_writer_slot has the same shape. Both now read the marker under the state lock and unlink only when the stored token is still ours, which is the same token re-check the stale-break path already relies on.

### DIFF
--- a/src/filelock/_soft_rw/_sync.py
+++ b/src/filelock/_soft_rw/_sync.py
@@ -386,7 +386,7 @@ class SoftReadWriteLock(metaclass=_SoftRWMeta):
         # enough (a stop-the-world GC pause, SIGSTOP, a suspended VM) for a peer to evict the marker as
         # stale and claim the writer slot itself, the file now at <path>.write is the peer's live marker;
         # unlinking it by path would let a second writer through and break mutual exclusion. The state lock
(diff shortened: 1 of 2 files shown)


src/filelock/_soft_rw/_sync.py
```diff
@@ -379,7 +379,23 @@ def release(self, *, force: bool = False) -> None:
         if hold.is_reader:
```
gaborbernat, line 379: src/filelock/_soft_rw/_sync.py:389: serialises ==> serializes

gaborbernat (changes requested):

src/filelock/_soft_rw/_sync.py
dxbjavid: good catch, fixed - switched it to serializes to match the rest of the tree.

gaborbernat (approved):

dxbjavid:
Thanks

PR #571, opened 2026-06-26
don't complete a writer acquire on a peer's reclaimed marker

dxbjavid:
the writer acquire is two-phase, and in phase 2 (waiting for readers to drain) it refreshes the .write marker on every scan. if the writer is paused longer than stale_threshold a peer can evict that now-stale marker and reclaim the slot with its own token, but phase 2 carried on touching whatever sat at the path and reported success once the readers had gone, so the original writer would finish its acquire on top of the peer's live marker and two writers could end up believing they hold the exclusive lock at the same time. this re-checks that the marker still carries our token before refreshing it, the same identity check the heartbeat refresh and the release path already rely on, and re-claims the slot (waiting behind the peer when it currently holds .write) instead of trusting a foreign marker. the regression test drives the eviction mid phase-2 and asserts the acquire no longer completes on a stranger's marker.

### DIFF
--- a/src/filelock/_soft_rw/_sync.py
+++ b/src/filelock/_soft_rw/_sync.py
@@ -397,6 +397,40 @@ class SoftReadWriteLock(metaclass=_SoftRWMeta):
                 return
             _unlink(self._paths.write)
 
+    def _claim_writer_marker(self, token: str) -> bool:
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):
LGTM

dxbjavid:
Thanks.

PR #576, opened 2026-07-02
use a private break name in break_lock_file

dxbjavid:
break_lock_file renames a stale lock aside before re-checking and unlinking it, but the aside name is a fixed `<lock>.break.<pid>`. Two breakers in the same process share that name (a thread pool where each worker holds its own lock instance with a lifetime, all contending on one expired lock), so a second break can rename a freshly recreated live lock onto that path in the small window between the re-verify lstat and the unlink, and the first breaker then deletes a live lock the inode check just approved, letting a second holder in. The soft read/write marker break already appends a random token to its aside name for exactly this reason; this makes break_lock_file do the same so no peer can target our break path.

### DIFF
--- a/src/filelock/_util.py
+++ b/src/filelock/_util.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import secrets
 import stat
 import sys
 from errno import EACCES, EISDIR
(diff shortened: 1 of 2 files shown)


gaborbernat (approved):
LGTM

dxbjavid:
Thank you @gaborbernat

PR #577, opened 2026-07-02
Keep Unix lock files after release

### LINKED ISSUE #574: UnixFileLock deleting the lock file on release can break mutual exclusion
## Summary

Since version 3.20.4, `UnixFileLock._release()` deletes the lock file after releasing. This breaks use cases where multiple processes coordinate via the same lock file path, because a process that already has an open fd on the old inode can hold a lock on a deleted file while another process creates and locks a new file under the same name.

## The issue

Citing the comment in https://github.com/tox-dev/filelock/issues/27#issuecomment-536073252, which is precisely the issue I faced.
> Deleting lockfiles on UNIX causes a race condition. Consider the following set of events:
>
>   - Program A creates and opens (as an atomic operation) a file A.lck.
>   - Program A grabs a lock on the file handle it holds on that file; this operation succeeds.
>   - Program B opens A.lck.
>   - Program B requests a lock on the file handle; this operation blocks.
>   -  Program A deletes A.lck.
>   - Program A closes its file handle, which makes program B able to get a lock on the file handle it already holds, on the now-deleted copy of A.lck.
>   - Program C creates and opens (as an atomic operation) a new file under the name A.lck.
>   - Program C grabs a lock on the file handle it holds on this file; because they're two different files (even if under the same name), this succeeds.
>
> ...thus, you have both programs B and C thinking they hold the same lock.

## Reproduction

Tested locally (Darwin, Python 3.12):

```python
# /// script
# requires-python = ">=3.12"
# dependencies = ["filelock"]
# ///
import platform, tempfile
from pathlib import Path
import filelock

p = Path(tempfile.gettempdir()) / "filelock_delete_test.lock"
p.unlink(missing_ok=True)
lock = filelock.FileLock(p)
lock.acquire()
lock.release()
print(
    f"filelock=={filelock.__version__}  "
    f"class={type(lock).__name__}  "
    f"platform={platform.system()}  "
    f"lock_file_survives_release={p.exists()}"
)
p.unlink(missing_ok=True)
```

Run across versions:
```bash
for v in 3.20.3 3.20.4 3.29.4; do
  uv run --with "filelock==$v" test_filelock_race.py
done
```

```raw
filelock==3.20.3  lock_file_survives_release=True
filelock==3.20.4  lock_file_survives_release=False  ← regression
filelock==3.29.4  lock_file_survives_release=False
```

## Suggested fix

We had to replace `filelock` with a raw `fcntl.flock`-based lock that intentionally leaves the lock file on disk. Ideally, `UnixFileLock._release()` should not delete the lock file, or have an option to keep it. This was the behavior through 3.20.3 and matches the semantics of `flock()` on UNIX. The lock file is inert after release (zero bytes, no sensitive content) and can be safely left on disk.

itscloud0:
## Problem

Since `03b0ab7` (`feat(unix): delete lock file on release`), `UnixFileLock._release()` unlinks the path before unlocking and closing the fd. That lets a process that already opened the old inode keep waiting on that deleted file while another process recreates the filename and locks a new inode, so both can believe they hold the same lock.

Fixes #574.

## Solution

- stop unlinking the Unix lock file on release
- keep the Unix-specific expectations aligned with the persistent-path behavior
- add a regression that proves a waiter fd cannot acquire a different inode after another lock instance reacquires the same path

## Testing

- `.venv-ci/bin/python -m pytest tests/test_filelock.py tests/test_unix_fallback.py tests/test_self_deadlock.py -q`
- `.venv-ci/bin/python -m ruff check src/filelock/_unix.py tests/test_filelock.py tests/test_unix_fallback.py tests/test_self_deadlock.py`
- `.venv-ci/bin/python -m ruff format --check src/filelock/_unix.py tests/test_filelock.py tests/test_unix_fallback.py tests/test_self_deadlock.py`

### DIFF
--- a/src/filelock/_unix.py
+++ b/src/filelock/_unix.py
@@ -38,9 +38,9 @@ else:  # pragma: win32 no cover
         """
         Uses the :func:`fcntl.flock` to hard lock the lock file on unix systems.
 
-        Lock file cleanup: Unix and macOS delete the lock file reliably after release, even in
-        multi-threaded scenarios. Unlike Windows, Unix allows unlinking files that other processes
-        have open.
+        The lock file is intentionally left in place after release. Unlinking a locked file on Unix
+        can split waiters across different inodes and break mutual exclusion for processes that
+        coordinate via the same path.
         """
 
         def _acquire(self) -> None:  # noqa: C901, PLR0912
@@ -103,10 +103,8 @@ else:  # pragma: win32 no cover
         def _release(self) -> None:
             fd = cast("int", self._context.lock_file_fd)
             self._context.lock_file_fd = None
-            with suppress(OSError):
-                Path(self.lock_file).unlink()
             fcntl.flock(fd, fcntl.LOCK_UN)
-            with suppress(OSError):  # close can raise EIO on FUSE/Docker bind-mount filesystems after unlink
+            with suppress(OSError):  # close can raise EIO on FUSE/Docker bind-mount filesystems
                 os.close(fd)
 
 
--- a/tests/test_filelock.py
+++ b/tests/test_filelock.py
@@ -945,7 +945,7 @@ def test_mtime_zero_exit_branch(
         lock.acquire(timeout=0)
 
 
-@pytest.mark.parametrize("lock_type", [FileLock, SoftFileLock])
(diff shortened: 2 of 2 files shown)


gaborbernat (approved):
LGTM

PR #579, opened 2026-07-02
use a private break name in break_lock_file

gaborbernat:
Supersedes #576 — that PR had maintainer edits disabled on the fork, so I could not push the follow-up. This keeps @dxbjavid's original commit and adds a small test tidy on top.

`break_lock_file` renamed a stale lock aside to a fixed `<lock>.break.<pid>` before re-checking and unlinking it. Two breakers in the same process share that name, so a second break can rename a freshly recreated live lock onto that path in the window between the re-verify `lstat` and the `unlink`, and the first breaker then deletes a live lock the inode check just approved. This appends a random token to the break name (matching the soft read/write marker break) so no peer can target our break path.

### DIFF
--- a/src/filelock/_util.py
+++ b/src/filelock/_util.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import secrets
 import stat
 import sys
 from errno import EACCES, EISDIR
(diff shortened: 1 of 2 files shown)


gaborbernat:
Not needed — #576 was merged as 7595a7b while this was in flight. Folding the follow-up test tidy into a single PR against main instead.

PR #581, opened 2026-07-03
serialise singleton construction in FileLockMeta

dxbjavid:
FileLockMeta.__call__ reads and writes its singleton cache without a lock, so two threads constructing the same is_singleton=True lock for a fresh path both miss the cache and each build their own instance. That quietly breaks the singleton guarantee callers lean on for reentrant locking across instances, so a later reentrant acquire on what should be the same lock can hit the deadlock check in acquire(), and the concurrent writes to the shared WeakValueDictionary are a race in their own right. ReadWriteLock and SoftReadWriteLock already take their _instances_lock around the get-or-create, so this brings the base metaclass in line.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -9,7 +9,7 @@ import time
 import warnings
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
-from threading import local
+from threading import Lock, local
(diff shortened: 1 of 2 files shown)


dxbjavid:
TY!

gaborbernat:
Thanks

PR #595, opened 2026-07-12
✨ feat(api): add optional lock-file opener hook

### LINKED ISSUE #592: Allow `FileLock` callers to provide a backward-compatible file opener
Severity: Enhancement (security-sensitive callers currently need to subclass private acquisition methods)

## Summary

I would like `FileLock` to accept an optional opener callback for the lock-file
descriptor. The default must preserve all current behavior, while callers with
stronger local-filesystem requirements could atomically apply platform-specific
open flags and handle validation without replacing `_acquire()`.

Examples include opening with stricter no-follow semantics, applying a private
descriptor policy, using native Windows handle flags, or instrumenting opens for
fault-injection tests.

## Environment

- package/version: `filelock 3.29.7`
- source: commit `1efb8932c08789deb08ad93a91d8996cc36bb9cc`
- platforms: Linux, macOS, and Windows local filesystems
- runtime: supported CPython versions
- relevant API: `BaseFileLock.__init__()` and platform `_acquire()` methods

## Reproduction

1. Attempt to configure the descriptor-opening operation on `FileLock`.
2. Observe that the constructor exposes path, mode, timeout, blocking,
   singleton, polling, and lifetime controls, but no opener hook.
3. Subclassing requires replacing platform-private `_acquire()` behavior,
   including contention classification and fallback behavior.

## Observed vs expected

Observed:

```text
Changing only the open operation requires duplicating a private platform
acquisition method.
```

Expected:

```text
An optional opener hook can replace descriptor creation while the library
continues to own polling, locking, release, and error classification.
```

## Root cause (verified in source, 3.29.7)

The constructor in
[`src/filelock/_api.py:208-260`](https://github.com/tox-dev/filelock/blob/1efb8932c08789deb08ad93a91d8996cc36bb9cc/src/filelock/_api.py#L208-L260)
has no opener contract. Unix acquisition calls `os.open()` directly in
[`src/filelock/_unix.py:46-75`](https://github.com/tox-dev/filelock/blob/1efb8932c08789deb08ad93a91d8996cc36bb9cc/src/filelock/_unix.py#L46-L75),
and Windows does the same in
[`src/filelock/_windows.py:59-86`](https://github.com/tox-dev/filelock/blob/1efb8932c08789deb08ad93a91d8996cc36bb9cc/src/filelock/_windows.py#L59-L86).

## Suggested fix

Add a keyword-only optional callback with a narrow descriptor-returning
contract, for example:

```python
FileOpener = Callable[[str, int, int], int]

def __init__(..., *, opener: FileOpener | None = None) -> None:
    ...
```

When `opener is None`, use the existing platform behavior unchanged. When it is
provided, call it with the final path, flags, and mode, then retain all existing
library-owned lock and cleanup logic. Validate that the result is a usable
integer descriptor. Include `opener` in singleton compatibility checks.

If a cross-platform callback is too broad, a smaller POSIX-only opener hook
would still remove the most common need for private-method duplication.

## Verification

- Existing callers observe no signature or runtime behavior change.
- A callback receives the expected path, flags, and mode exactly once per
  attempt and its descriptor is closed on contention.
- Callback exceptions propagate without being reported as timeout.
- Singleton construction rejects incompatible opener identities.
- Sync and async wrappers use the same configured opener safely.

## Current workaround

Subclass the platform lock and replace `_acquire()`, or place the lock file in a
prevalidated owner-private directory. The subclass approach duplicates private
behavior and is sensitive to upstream changes.

gaborbernat:
Callers with stronger local-filesystem requirements must subclass a private platform `_acquire` just to change how the lock-file descriptor opens, which couples them to internals like contention classification and fallback behavior, as raised in #592. Examples include applying stricter no-follow semantics, a private-descriptor policy, native Windows handle flags, or instrumenting opens for fault injection.

This adds a keyword-only `opener` callback with the narrow contract `(path, flags, mode) -> fd`. The callback threads through the constructor, the lock context, and the singleton compatibility check, and every backend calls it in place of its direct `os.open`. When `opener` is `None`, which is the default, each backend opens as before. When a caller provides one, the library still owns polling, locking, release, and error classification; the callback only creates the descriptor. The package exports `FileOpener` for callers to type their hooks.

An opener that returns anything other than a non-negative integer descriptor raises `ValueError`, and an exception from inside the callback propagates instead of surfacing as a timeout. Singleton construction treats a differing opener identity as an incompatible argument, and the sync and async wrappers share the same configured opener.

Closes #592.

### DIFF
--- a/src/filelock/__init__.py
+++ b/src/filelock/__init__.py
@@ -12,7 +12,7 @@ import sys
 import warnings
 from typing import TYPE_CHECKING, Final
 
-from ._api import AcquireReturnProxy, BaseFileLock
+from ._api import AcquireReturnProxy, BaseFileLock, FileOpener
 from ._error import Timeout
 
 if TYPE_CHECKING:
@@ -82,6 +82,7 @@ __all__ = [
     "BaseAsyncFileLock",
     "BaseFileLock",
     "FileLock",
+    "FileOpener",
     "ReadWriteLock",
     "SoftFileLock",
     "SoftReadWriteLock",
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -10,7 +10,7 @@ import warnings
 from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass
 from threading import Lock, local
-from typing import TYPE_CHECKING, Any, Final, TypeVar
+from typing import TYPE_CHECKING, Any, Final, Protocol, TypeVar
 from weakref import WeakValueDictionary
 
 from ._error import Timeout
@@ -19,6 +19,22 @@ from ._util import break_lock_file
 #: No explicit file permission mode was passed. Lock files then open with 0o666 so umask and default ACLs pick
 #: the final permissions, and fchmod is skipped to preserve POSIX default ACL inheritance.
 _UNSET_FILE_MODE: Final[int] = -1
+
+
+class FileOpener(Protocol):
+    """Opener for the lock-file descriptor, standing in for the ``os.open`` call a backend would otherwise make."""
+
+    def __call__(self, path: str, flags: int, mode: int, /) -> int:
+        """:returns: an open file descriptor for *path*, opened with *flags* and *mode*."""
+
+
+class _OpenerFailed(Exception):  # noqa: N818
+    """Carries an opener's own exception past backend contention handling so acquire can re-raise it unchanged."""
+
+    def __init__(self, cause: BaseException) -> None:
+        super().__init__()
+        self.cause = cause
+
 
 if TYPE_CHECKING:
     from collections.abc import Callable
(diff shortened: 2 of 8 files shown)


gaborbernat:
Closing per review. The `_OpenerFailed` wrapper makes an ordinary `FileExistsError` escape instead of reporting contention, a callback's first result can influence later attempts in a way that bypasses permission and file-type policy, and on Windows the callback cannot supply the validated raw handle that the reparse-point fix (#596) requires. A constrained POSIX-only design could revive this later: replace only the `os.open`, keep validation and cleanup in filelock, let non-`OSError` callback exceptions propagate, and treat `FileExistsError` as contention only when it comes from exclusive marker creation.

PR #596, opened 2026-07-12
🐛 fix(windows): bind reparse-point check to the locked handle

gaborbernat:
The Windows acquisition path validated the lock pathname with `GetFileAttributesW` to reject reparse points, then opened the same pathname in a separate `os.open` that follows reparse points. Nothing bound the check and the open to one handle, so an attacker able to write the lock directory could replace the path with a symlink or junction in the window between them and redirect the lock to another file. A comment claimed the check prevented a symlink TOCTOU, but nothing tied the decision to the descriptor the lock used.

This opens the path with `CreateFileW` and `FILE_FLAG_OPEN_REPARSE_POINT` so a symlink is not followed, reads the reparse attribute from that same handle through `GetFileInformationByHandle`, rejects a reparse point, and only then wraps the handle into a descriptor for `msvcrt.locking`. The security decision now comes from the handle the lock takes rather than a prior pathname query, so a path swapped after the check no longer changes what gets locked. The backend still creates a missing lock file, and the open still treats access or sharing errors as contention.

The change stays inside the Windows backend and preserves the existing acquire, release, and contention semantics. Placing lock files in a directory untrusted users cannot modify remains the strongest mitigation, since it removes the attacker-controlled path-replacement precondition.

### DIFF
--- a/src/filelock/_windows.py
+++ b/src/filelock/_windows.py
@@ -15,44 +15,77 @@ if sys.platform == "win32":  # pragma: win32 cover
     import msvcrt
     from ctypes import wintypes
 
+    _GENERIC_READ: Final[int] = 0x80000000
+    _GENERIC_WRITE: Final[int] = 0x40000000
+    _FILE_SHARE_READ_WRITE: Final[int] = (
+        0x00000001 | 0x00000002
+    )  # read | write; matches os.open (_SH_DENYNO), no delete
+    _OPEN_ALWAYS: Final[int] = 4  # open the file if it exists, create it otherwise
+    _FILE_ATTRIBUTE_READONLY: Final[int] = 0x00000001
     _FILE_ATTRIBUTE_REPARSE_POINT: Final[int] = 0x00000400
-    _INVALID_FILE_ATTRIBUTES: Final[int] = 0xFFFFFFFF
+    _FILE_FLAG_OPEN_REPARSE_POINT: Final[int] = 0x00200000  # open the reparse point itself instead of following it
+    _INVALID_HANDLE_VALUE: Final[int] = cast("int", wintypes.HANDLE(-1).value)  # non-null handle, never None
+    _ERROR_ACCESS_DENIED: Final[int] = 5
+    _ERROR_SHARING_VIOLATION: Final[int] = 32
+    _OWNER_WRITE: Final[int] = 0o200
 
     _kernel32: Final[ctypes.WinDLL] = ctypes.WinDLL("kernel32", use_last_error=True)
-    _kernel32.GetFileAttributesW.argtypes = [wintypes.LPCWSTR]
-    _kernel32.GetFileAttributesW.restype = wintypes.DWORD
+    _kernel32.CreateFileW.argtypes = [
+        wintypes.LPCWSTR,
+        wintypes.DWORD,
+        wintypes.DWORD,
+        wintypes.LPVOID,
+        wintypes.DWORD,
+        wintypes.DWORD,
+        wintypes.HANDLE,
+    ]
+    _kernel32.CreateFileW.restype = wintypes.HANDLE
+    _kernel32.CloseHandle.argtypes = [wintypes.HANDLE]
+    _kernel32.CloseHandle.restype = wintypes.BOOL
+
+    class _BY_HANDLE_FILE_INFORMATION(ctypes.Structure):  # noqa: N801  # mirrors the Win32 struct name
+        _fields_ = (
+            ("dwFileAttributes", wintypes.DWORD),
+            ("ftCreationTime", wintypes.FILETIME),
+            ("ftLastAccessTime", wintypes.FILETIME),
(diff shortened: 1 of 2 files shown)


mmashwani:
Severity: Unreleased integration blocker (the approved opener contract is silently ignored by the approved secure Windows backend)

## Summary

I combined the approved branches for PR #595 and PR #596 and found an
integration conflict. PR #595 adds an `opener` callback and states that every
backend calls it in place of direct descriptor creation. PR #596 replaces the
Windows open path with a module-level `_open_non_reparse_fd()` helper that calls
`CreateFileW()` directly. The configured opener is therefore never invoked on
Windows when both changes are present.

The security change should remain handle-bound. The integration should not
silently discard a configured public callback.

This is not just a callback plumbing issue. On Windows, the security boundary is
the handle returned by the backend open operation. A generic `os.open`-style
callback cannot, by itself, prove that the opened handle preserved
`FILE_FLAG_OPEN_REPARSE_POINT`, inspected the same handle it will lock, and
retained the intended sharing semantics. If the opener contract is kept for
Windows, the contract needs to express those handle-bound requirements directly.

## Environment

- public baseline: `filelock 3.29.7`, commit
  `1efb8932c08789deb08ad93a91d8996cc36bb9cc`
- opener branch: PR #595, commit `2e98912`
- secure Windows branch: PR #596, commit `8f9999e`
- local integration branch: `analysis/approved-lock-fixes-current` at `0ed99ca`
- local integration order: current `main`, then #593, #594, #595, and #596
- platform contract: Windows local filesystems

## Reproduction

1. Combine the head commits from PR #595 and PR #596.
2. Run the attached
   `test_windows_opener_secure_branch_integration.py` against that combined
   source.
3. The test configures an opener and expects one call during physical
   acquisition.
4. Observe that the assertion fails because the call count is zero.

## Observed vs expected

Observed:

```text
WindowsFileLock accepts opener=... through the shared constructor but bypasses
it and opens through CreateFileW directly.
```

Expected:

```text
The combined API either honors a secure Windows-capable opener contract or
rejects an unsupported opener explicitly. It must not silently ignore it.
```

## Root cause (verified in the approved branches)

PR #595 routes Windows descriptor creation through
[`self._open()`](https://github.com/tox-dev/filelock/blob/2e9891251b649a70de40fdfdaf2efd3962bcd0e6/src/filelock/_windows.py#L42-L55).

PR #596 changes acquisition to call
[`_open_non_reparse_fd()`](https://github.com/tox-dev/filelock/blob/8f9999e249388853e711db0509a10cbf4ad5b71b/src/filelock/_windows.py#L75-L91),
whose
[`CreateFileW()` implementation](https://github.com/tox-dev/filelock/blob/8f9999e249388853e711db0509a10cbf4ad5b71b/src/filelock/_windows.py#L102-L165)
has no lock instance or opener parameter.

The branches are individually coherent but modify the same descriptor-opening
boundary in incompatible ways.

## Suggested fix

Resolve the public contract before either branch is released. Safe options are:

1. Add a Windows-specific opener protocol capable of preserving
   `FILE_FLAG_OPEN_REPARSE_POINT`, the share-mode policy, and handle-bound
   validation on the same handle that will be locked.
2. Keep the secure backend-owned opener and reject `opener` explicitly on
   Windows until a safe callback contract exists.
3. Redesign the callback as an observation or fault-injection boundary around
   backend-owned secure opening instead of a generic replacement.

Routing an arbitrary `os.open`-style callback without a security contract could
reintroduce the reparse-point weakness that PR #596 fixes, so simply restoring
the PR #595 call is not sufficient.

If maintainers want to support security-sensitive lock paths beneath
less-trusted parent directories later, that should be a separate documented
contract because final-component no-follow does not validate intermediate
directory components.

## Verification

- A configured Windows opener is invoked exactly as the final documented
  contract specifies, or construction fails explicitly.
- Default Windows acquisition retains handle-bound reparse-point validation.
- A callback cannot silently downgrade the default no-follow behavior.
- Callback failures close every acquired handle and propagate without timeout
  misclassification.
- Sync, async, singleton, and contention tests cover the combined branches.

## Current workaround

Do not combine the two public contracts without an integration patch. A caller
cannot detect from successful construction that the configured opener will be
ignored on Windows.

## Regression test

<details>
<summary>test_windows_opener_secure_branch_integration.py</summary>

```python
"""Regression test for the combined custom-opener and secure Windows branches."""

from __future__ import annotations

import os
import sys
from pathlib import Path

import pytest


@pytest.mark.skipif(sys.platform != "win32", reason="Windows backend integration")
def test_windows_backend_uses_configured_opener(tmp_path: Path) -> None:
    from filelock import FileLock

    calls: list[tuple[str, int, int]] = []

    def opener(path: str, flags: int, mode: int) -> int:
        calls.append((path, flags, mode))
        return os.open(path, flags, mode)

    lock = FileLock(tmp_path / "resource.lock", timeout=0, opener=opener)
    with lock:
        assert lock.is_locked

    assert len(calls) == 1
```

</details>

Issue #601, opened 2026-07-13
`SoftFileLock` can create overlapping holders after malformed metadata publication


mmashwani:
Severity: High (malformed holder metadata can break mutual exclusion)

## Summary

I reproduced overlapping `SoftFileLock` holders when the PID and hostname record
is not fully published. Acquisition suppresses a metadata write error, and it
also ignores the byte count from a short `os.write()` call. In both cases, the
lock object records the descriptor as successfully acquired even though the
published lock file is malformed. Once that malformed file is old enough for
self-healing, a contender removes it and acquires a new inode while the original
lock object still reports ownership. The reproduction also verifies that the
original holder and contender retain descriptors for distinct files, and that
the lock pathname now points to the contender's file.

## Environment

- package/version: `filelock 3.29.7`
- source: commit `1efb8932c08789deb08ad93a91d8996cc36bb9cc`
- reproduced platform: macOS, local APFS filesystem
- reproduction method: deterministic `os.write()` error and short-write fault
  injection
- affected API: `SoftFileLock`

## Reproduction

1. Run the attached `filelock_soft_metadata_write_overlap.py` with
   `filelock==3.29.7`.
2. Observe:

   ```text
   filelock_version=3.29.7
   write_error.malformed_size=0
   write_error.holder_locked=True
   write_error.contender_locked=True
   write_error.overlap=True
   write_error.holder_fd_still_open=True
   write_error.descriptors_distinct=True
   write_error.path_matches_contender=True
   short_write.malformed_size=1
   short_write.holder_locked=True
   short_write.contender_locked=True
   short_write.overlap=True
   short_write.holder_fd_still_open=True
   short_write.descriptors_distinct=True
   short_write.path_matches_contender=True
   ```

The script ages the malformed file past the library's own malformed-file grace
period so the result does not depend on wall-clock waiting.

## Observed vs expected

Observed:

```text
A failed or partial metadata write is treated as successful acquisition. A later
contender reclaims the malformed file while the original holder remains live on
the old descriptor, so the two lock objects now refer to different files for the
same logical lock path.
```

Expected:

```text
Acquisition must fail closed if required holder metadata cannot be written
completely. A write failure or short write must never publish a reclaimable live
lock.
```

The holder record is part of the coordination state, not optional diagnostics.
If peers use malformed holder records as evidence that a lock can be reclaimed,
then the acquiring process should not record local ownership until it has
published a complete holder record that peers can parse under the same
protocol.

[...]

Issue #602, opened 2026-07-13
`SoftReadWriteLock` can create overlapping holders after a short marker write


mmashwani:
Severity: Mutual-exclusion break in the soft read/write lock when marker creation is partially written

## Summary

`SoftReadWriteLock` writes reader and writer marker payloads with one
`os.write()` call and does not verify that the full token/pid/hostname payload
was written. A short write can leave a malformed reader marker while
`acquire_read()` still reports success.

The holder's heartbeat later stops because it cannot parse its own marker. Once
the malformed marker becomes stale, a writer can break it and acquire the same
logical lock while the first lock object still believes it holds a read lock.

## Environment

- affected release: `filelock 3.29.7`
- source path: `src/filelock/_soft_rw/_sync.py`
- reproduced on: macOS with Python 3.12

## Reproduction

1. Install `filelock==3.29.7`.
2. Run the attached `filelock_soft_rw_short_marker_write_overlap.py`.
3. The script patches only the marker-payload `os.write()` call inside
   `filelock._soft_rw._sync` so the first marker write publishes only one byte
   of the marker payload.
4. The first lock acquires a read lock, then a second lock acquires a write lock
   after the malformed reader marker becomes stale.

Observed output:

```text
filelock_version=3.29.7
injected_short_write=True
malformed_marker_size=1
reader_marker_removed=True
writer_marker_exists=True
holder_still_thinks_read_locked=True
holder_heartbeat_alive=False
contender_write_locked=True
overlap=True
```

## Observed vs expected

Observed:

```text
The reader acquisition succeeds even though the reader marker payload was not
written. A writer can later acquire while the reader object still has a live
hold record.
```

Expected:

```text
Marker creation either writes the complete token/pid/hostname payload and
acquisition succeeds, or the partial marker is removed and acquisition fails
without creating a logical holder.
```

The marker file is the published ownership record for other processes and for
the holder's own heartbeat. Recording a local `_Hold` before that marker is
parseable creates split-brain state: the local object believes it holds the
lock, while peer-visible state says the marker is malformed and later
breakable.

## Root cause (verified in source, 3.29.7)

`_atomic_create_marker()` creates the marker file and calls `os.write(fd,
content)`, then closes the descriptor without checking the returned byte count:

https://github.com/tox-dev/filelock/blob/1efb8932c08789deb08ad93a91d8996cc36bb9cc/src/filelock/_soft_rw/_sync.py#L753-L765

[...]

PR #611, opened 2026-07-13
fail closed when a soft lock's holder record is only partially written

dxbjavid:
SoftFileLock._write_lock_info suppressed write errors, and neither it nor the read/write lock's _atomic_create_marker checked the byte count os.write returns, so a failed or short write leaves the pid/hostname record (or the token/pid/hostname marker) empty or truncated while acquire still records the descriptor as held. That published file parses back as malformed rather than a live holder, so once it ages past the malformed-lock grace threshold a contender breaks it and takes a fresh inode while the original object still owns its descriptor, and you end up with two overlapping holders on the same logical lock; for SoftReadWriteLock the holder's own heartbeat also cannot parse its marker and quietly stops. I reproduced both with fault injection on os.write on macos, closing issues #601 and #602. The holder record is coordination state rather than diagnostics, so the fix writes it in full or not at all: a small write_all helper loops over short writes and raises on error, SoftFileLock drops its just-created O_EXCL file and lets the poll loop retry, and _atomic_create_marker unlinks the partial marker and fails the claim.

### DIFF
--- a/src/filelock/_soft.py
+++ b/src/filelock/_soft.py
@@ -11,7 +11,7 @@ from pathlib import Path
 from typing import Final
 
 from ._api import BaseFileLock
-from ._util import break_lock_file, ensure_directory_exists, raise_on_not_writable_file
+from ._util import break_lock_file, ensure_directory_exists, raise_on_not_writable_file, write_all
 
 _WIN_SYNCHRONIZE: Final[int] = 0x100000
 _WIN_ERROR_INVALID_PARAMETER: Final[int] = 87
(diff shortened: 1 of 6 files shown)


gaborbernat:
Thanks for digging into this and reproducing it with fault injection on `os.write`. The same partial/short-write hole in the soft holder record was fixed in #614: it adds a write-all-or-nothing helper, drops the just-created `O_EXCL` file so the poll loop retries when a `SoftFileLock` write fails, and unlinks a partial marker before failing the claim in the read/write lock. That closed #601 and #602, so this change is superseded by the merged work. Appreciate the thorough writeup.

PR #612, opened 2026-07-13
fix(windows): propagate lock file open denials

### LINKED ISSUE #604: `WindowsFileLock` reports permanent open denial as lock timeout
Severity: Medium (a configuration or ACL failure is misclassified as lock contention)

## Summary

I found that `WindowsFileLock` treats every `EACCES` from `os.open()` as a
failed acquisition attempt. The polling loop then reports `Timeout`, even when
the denial is permanent and no competing lock holder exists.

This makes operational diagnosis difficult and can cause callers to retry a
permission failure as though the resource were busy.

For callers that treat lock acquisition as part of a fail-closed filesystem
operation, this distinction matters. A polling timeout is appropriate for a
contended byte-range lock or a transient sharing violation. It is not a good
contract for a path, ACL, or policy failure that no amount of waiting can fix.

## Environment

- package/version: `filelock 3.29.7`
- source: commit `1efb8932c08789deb08ad93a91d8996cc36bb9cc`
- platform: Windows
- reproduction method: deterministic `os.open()` fault injection
- affected APIs: `FileLock`, `WindowsFileLock`, async Windows wrappers

## Reproduction

1. Run the attached `filelock_windows_access_denied_timeout.py` with
   `filelock==3.29.7` on Windows.
2. The script injects `PermissionError(EACCES)` from the lock-file open.
3. Observe:

   ```text
   filelock_version=3.29.7
   injected_exception=PermissionError
   propagated_exception=Timeout
   observed_is_timeout=True
   ```

## Observed vs expected

Observed:

```text
A permanent open denial is swallowed and becomes Timeout.
```

Expected:

```text
File-open permission failures propagate as permission errors. Only a native
locking conflict, or a documented transient sharing condition, is classified as
contention and timeout.
```

## Root cause (verified in source, 3.29.7)

[`WindowsFileLock._acquire()`](https://github.com/tox-dev/filelock/blob/1efb8932c08789deb08ad93a91d8996cc36bb9cc/src/filelock/_windows.py#L59-L86)
uses the same `EACCES` classification for two different operations:

- `os.open()` denial at lines 73-77
- `msvcrt.locking()` contention at lines 79-84

The approved secure Windows implementation in PR #596 retains the same
classification by returning contention for both `ERROR_ACCESS_DENIED` and
`ERROR_SHARING_VIOLATION`.

Microsoft documents these as distinct system errors: `ERROR_ACCESS_DENIED`
means access is denied, while `ERROR_SHARING_VIOLATION` means another process is
using the file:
<https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499->.

## Suggested fix

Separate permanent open failures from lock conflicts:

- Propagate `EACCES` or `ERROR_ACCESS_DENIED` from descriptor creation unless a
  narrower documented Windows condition proves it is transient contention.
- Continue treating `EACCES` from `msvcrt.locking()` and
  `ERROR_SHARING_VIOLATION` from `CreateFileW()` as contention.
- Preserve the original OS error code, operation, and path for permanent
  failures.
- Prefer making this classification before the polling loop records a failed
  attempt, so permanent acquisition failures do not become indistinguishable
  from a real timeout.

## Verification

- Inject permanent `ERROR_ACCESS_DENIED` during open and assert it propagates.
- Inject `ERROR_SHARING_VIOLATION` and assert polling and timeout remain.
- Inject `EACCES` from `msvcrt.locking()` and assert it remains contention.
- Cover immediate timeout, finite timeout, sync, and async APIs.

## Current workaround

Preflight the lock directory and file ACLs before acquisition, then interpret a
Windows timeout alongside separate permission diagnostics. A preflight is
still path-based and cannot classify the actual open atomically.

## Reproduction script

<details>
<summary>filelock_windows_access_denied_timeout.py</summary>

```python
"""Show that a permanent Windows open denial is reported as lock contention."""

from __future__ import annotations

import errno
import sys
import tempfile
from pathlib import Path
from unittest.mock import patch


def main() -> int:
    if sys.platform != "win32":
        print("This reproduction targets WindowsFileLock.")
        return 2

    import filelock
    from filelock import FileLock, Timeout
    from filelock import _windows as windows_backend

    with tempfile.TemporaryDirectory() as directory:
        path = Path(directory) / "resource.lock"
        observed: BaseException | None = None

        denial = PermissionError(errno.EACCES, "injected permanent access denial", str(path))
        with patch.object(windows_backend.os, "open", side_effect=denial):
            try:
                FileLock(path, timeout=0).acquire()
            except BaseException as exc:  # capture the public exception classification
                observed = exc

        propagated = type(observed).__name__ if observed else None
        observed_is_timeout = isinstance(observed, Timeout)
        print(f"filelock_version={filelock.__version__}")
        print(f"injected_exception={type(denial).__name__}")
        print(f"propagated_exception={propagated}")
        print(f"observed_is_timeout={observed_is_timeout}")
        return 0 if observed_is_timeout else 1


if __name__ == "__main__":
    raise SystemExit(main())
```

</details>

Sanjays2402:
## Summary

- propagate permanent `os.open()` permission failures from `WindowsFileLock`
- keep treating `EACCES` from `msvcrt.locking()` as byte-range lock contention
- add a platform-simulated regression test that exercises the Windows backend on any CI host

Fixes #604.

## Testing

- `pytest -q tests/test_windows.py` — 1 passed
- `pytest -q tests/test_windows.py tests/test_filelock.py` — 145 passed, 20 skipped before a formatter-only change
- changed-file Ruff lint and format checks pass

### DIFF
--- a/src/filelock/_windows.py
+++ b/src/filelock/_windows.py
@@ -39,20 +39,15 @@ if sys.platform == "win32":  # pragma: win32 cover
                 msg = f"Lock file is a reparse point (symlink/junction): {self.lock_file}"
                 raise OSError(msg)
 
+            fd = os.open(self.lock_file, os.O_RDWR | os.O_CREAT, self._open_mode())
             try:
-                fd = os.open(self.lock_file, os.O_RDWR | os.O_CREAT, self._open_mode())
+                msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
             except OSError as exception:
-                if exception.errno != EACCES:
+                os.close(fd)
+                if exception.errno != EACCES:  # EACCES means another holder owns the byte-range lock
                     raise
             else:
-                try:
-                    msvcrt.locking(fd, msvcrt.LK_NBLCK, 1)
-                except OSError as exception:
-                    os.close(fd)
-                    if exception.errno != EACCES:  # EACCES means another holder owns the byte-range lock
-                        raise
-                else:
-                    self._context.lock_file_fd = fd
+                self._context.lock_file_fd = fd
 
         def _release(self) -> None:
             fd = cast("int", self._context.lock_file_fd)
--- a/tests/test_windows.py
+++ b/tests/test_windows.py
@@ -1,0 +1,51 @@
+from __future__ import annotations
+
+import ctypes
+import importlib
+import sys
+from errno import EACCES
+from typing import TYPE_CHECKING
+
+import pytest
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from pytest_mock import MockerFixture
+
+
+def test_open_permission_error_is_not_reported_as_contention(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    mocker: MockerFixture,
+) -> None:
+    """A permanent descriptor-open denial must escape before the polling loop."""
+    import filelock._windows as windows_module
+
+    fake_kernel32 = mocker.Mock()
+    fake_kernel32.GetFileAttributesW = mocker.Mock(return_value=0xFFFFFFFF)
+    fake_msvcrt = mocker.Mock()
+    fake_msvcrt.LK_NBLCK = 1
+    fake_msvcrt.LK_UNLCK = 0
+    fake_msvcrt.locking = mocker.Mock()
+
+    original_platform = sys.platform
+    monkeypatch.setitem(sys.modules, "msvcrt", fake_msvcrt)
+    monkeypatch.setattr(sys, "platform", "win32")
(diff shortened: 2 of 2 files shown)


gaborbernat:
Thanks for the fix and the cross-platform regression test. #604 was addressed in #617, which opens the lock file through `NtCreateFile` and reads the raw NTSTATUS so a delete-pending name, a sharing violation, and a real access denial are told apart instead of collapsing into one `ERROR_ACCESS_DENIED`. That closed #604, so this PR is superseded by the merged work. Thanks for taking the time on it.

PR #670, opened 2026-07-16
reject non-finite lease duration in marker records

dxbjavid:
_build_record parses a marker's lease duration with float(), which also accepts "nan" and "inf"; the following duration <= 0 guard lets both through, since neither compares as non-positive. A peer sharing the filesystem that plants a lease marker carrying duration=nan is then read as a valid claim rather than a malformed one, so every contender raises LeaseSettingsMismatch (nan compares unequal to any configured duration) and the stale marker is never reclaimed, wedging acquisition on that path. Rejecting a non-finite duration lets such a marker fall through to the malformed path and self-heal, the same way an out-of-range pid already does.


dxbjavid:
TY

PR #681, opened 2026-07-20
🐛 fix(strict): tolerate an errno without ENOTSUP

gaborbernat:
`_strict.py` imports `ENOTSUP` from `errno` at module scope. GraalPy's `errno` does not define it, so importing `filelock` on that runtime raises:

```
ImportError: cannot import name 'ENOTSUP' from 'errno'
```

This breaks every consumer on GraalPy at import, before anyone takes a lock. pypa/virtualenv's GraalPy jobs die at conftest import on its `main`: [ubuntu-24.04](https://github.com/pypa/virtualenv/actions/runs/29730396773/job/88313263808) and [macos-15](https://github.com/pypa/virtualenv/actions/runs/29730396773/job/88313263782). virtualenv requires `filelock>=3.24.2,<4`, so it resolves to a release carrying the import. `ENOTSUP` arrived with the claim protocol in #659 and shipped first in 3.30.0. GraalPy 24.1, 24.2 and 25.1 all omit the name, so this is not a single bad release.

## The fix

`_strict.py` now probes the code instead of importing it.

- `ENOTSUP` wins wherever it exists, so no platform that defines the name changes behavior.
- `EOPNOTSUPP` stands in only for runtimes that lack `ENOTSUP`. It approximates rather than matches: the two agree on Linux and differ on macOS/BSD (45 against 102) and on Windows (129 against 130).
- Where neither exists, the code drops out and `ENOSYS`/`EXDEV` still classify the link failures such a runtime can raise.

The probe follows the module's existing convention (`_probe_link_follow_symlinks`) and runs once at import.

## Four changes ride along

**A coverage measurement bug.** `BaseAsyncFileLock.__del__`'s no-loop path reported its `except RuntimeError` and the following `return` as missing while the handler body between them reported covered, which no execution can produce. coverage refuses its `sys.monitoring` core when a config sets contexts, since that core does not implement them. pytest-cov's `--cov-context` switches contexts through another path, behind that guard. Every 3.14 job had been running the combination coverage rules out, and the seed decided which lines went missing. Dropping `--cov-context` fixes it. Pinning the C tracer does not: Termux ships coverage without the extension.

**A GraalPy CI job.** Linux only, and unmeasured like PyPy, so the 100% gate stays CPython's. Nothing else checks the capability probes against a second runtime.

**Capability gating.** GraalPy ran the suite with 55 failures, none of them filelock defects. Each now gates on a probe of the behavior it needs. Three turned out to be real test bugs, so I fixed those rather than gating them. Twenty cancellation tests carry a non-strict xfail against a GraalPy `contextlib` that answers `athrow()` with `RuntimeError`.

**One gate registry.** Seven marks had been redefined per module, `fcntl` in six and `fork` in five, and three re-derived their answer inline instead of reading the probe. That last one hid a bug: GraalPy registers fork handlers without being able to fork, so a capability asking for both skipped four tests needing only the registration.

## Verification

| Runtime | Result |
| --- | --- |
| CPython 3.14 | 1239 passed, 52 skipped, 100% total and diff coverage |
| GraalPy 24.2.2 | 1105 passed, 153 skipped, 28 xfailed, 5 xpassed |
| PyPy 3.11 | 1226 passed, 65 skipped |

`ty` and `pre-commit` clean. `src/filelock` gains only the probe, and no pre-existing capability value changes on CPython, so it measures the lines it measured before.

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -27,12 +27,17 @@ jobs:
           - "3.11"
           - "3.10"
           - "pypy3.11"
+          - "graalpy3.11"
         os:
           - ubuntu-24.04
           - windows-2025
           - macos-15
         exclude:
           - {os: windows-2025, py: pypy3.11}
+          # GraalPy validates the capability probes the suite gates on, which is a per-OS answer, not a per-arch one.
+          # One Linux job settles it; the Windows and macOS builds only repeat it at triple the runtime.
+          - {os: windows-2025, py: graalpy3.11}
+          - {os: macos-15, py: graalpy3.11}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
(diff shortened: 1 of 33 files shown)


gaborbernat:
CI: the two red jobs are the flaky `src/filelock/asyncio.py` 683/686 coverage miss that also fails on `main` (run 29735645550, same lines). Every test passes in both.

PR #686, opened 2026-07-22
hand back a soft read/write claim when the caller is cancelled

dxbjavid:
AsyncSoftReadWriteLock dispatches the sync acquire through run_in_executor, which cannot recall work the pool has already started, so cancelling the awaiting task only detaches the caller: the acquire carries on, creates its reader or .write marker, sets the hold and starts the heartbeat. Nothing ever releases that claim, and since the heartbeat keeps refreshing the marker no peer on any host can even evict it as stale, so every contender is wedged until the process exits. AsyncReadWriteLock already reconciles this by draining the submitted call and handing the claim back, so this does the same here, drains a cancelled release or close rather than leaving its outcome unobserved, and moves the shared cancellation reporting into _async so both wrappers behave alike.


tests/soft_rw/test_soft_rw_async_cancellation.py
```diff
@@ -0,0 +1,176 @@
+from __future__ import annotations
```
gaborbernat: No new test file add it to existing.

gaborbernat (changes requested):

gaborbernat (approved):

dxbjavid:
TY!

PR #691, opened 2026-07-28
hand back the claim when a heartbeat thread fails to start

dxbjavid:
SoftReadWriteLock and SoftFileLease both commit the hold and its on-disk marker before their heartbeat thread has started, so if the OS refuses a new thread the marker is left with nothing refreshing it: a peer evicts it as stale and acquires while this instance still believes it holds the lock, and a later release then raises trying to join a thread that never started, so the marker is not even unlinked. Start the heartbeat under the same critical section that publishes the hold and, on a start failure, clear the hold and drop the marker we just claimed; the lease records its heartbeat only once the thread is running so the acquire rollback releases the claim cleanly. Both paths come with a regression test that fails on the current code.


dxbjavid:
TY!

Issue #701, opened 2026-08-12
Audit hook crashes during interpreter shutdown


vladimirnot:
Hello, highly esteemed maintainers!

We hit this error while shutting down Python workers:

```python
Exception ignored in audit hook:
Traceback (most recent call last):
  File ".../filelock/_api.py", line 1785, in _audit_fork_safety
TypeError: argument of type 'NoneType' is not a container or iterable
```

It looks like CPython clears the filelock._api globals before it removes the registered audit hook. A late audit event then calls _audit_fork_safety() after _FORK_AUDIT_EVENTS becomes None.

We saw this with Python 3.14 during ASGI worker teardown. An unclosed uvloop resource triggered id(self) while Python formatted its warning.

Here is a small reproduction:
```python
from collections.abc import Callable

import filelock._api as api


class LateAudit:
    def __del__(self, object_id: Callable[[object], int] = id) -> None:
        object_id(self)


api.__dict__["_zz_shutdown_audit"] = LateAudit()
```

PR #702, opened 2026-08-13
fix(api): handle module teardown NoneType in _audit_fork_safety during interpreter shutdown (#701)

harmehak0173:
### Summary of Changes
Fixes a `TypeError: argument of type 'NoneType' is not a container or iterable` crash in `_audit_fork_safety` during interpreter shutdown (#701).

### Root Cause
During CPython interpreter finalization/shutdown, module global variables (such as `_FORK_AUDIT_EVENTS` and `_FORK_STATE` in `filelock/_api.py`) are cleared and set to `None` before registered audit hooks are removed. When late audit events fire during teardown, calling `_audit_fork_safety` executes `if event in _FORK_AUDIT_EVENTS:`, which raises a `TypeError` because `_FORK_AUDIT_EVENTS` is `None`.

### Solution
1. Defensively check `if _FORK_AUDIT_EVENTS is not None and event in _FORK_AUDIT_EVENTS:` in `_audit_fork_safety`.
2. Check `if _FORK_STATE is not None` before accessing `transition_context` or `fork_owner_depths`.
3. Added unit test `test_audit_fork_safety_during_interpreter_shutdown` in `tests/test_fork_coordination.py`.

### Testing
- `ruff check` & `ruff format`: All checks passed.
- `pytest`: 1,301 passed.

### DIFF
--- a/src/filelock/_api.py
+++ b/src/filelock/_api.py
@@ -1782,11 +1782,11 @@ def _refresh_owner_pins() -> None:  # pragma: needs fork
 def _audit_fork_safety(  # pragma: no cover - CPython disables tracing while Python audit hooks run
     event: str, _args: tuple[object, ...]
 ) -> None:
-    if event in _FORK_AUDIT_EVENTS:
-        if _FORK_STATE.transition_context.depth or _FORK_STATE.fork_owner_depths:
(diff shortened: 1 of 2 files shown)


gaborbernat:
superseeded by https://github.com/tox-dev/filelock/pull/703

PR #709, opened 2026-08-17
escape the hostname every marker publishes

dxbjavid:
Every marker format in this package holds the hostname to printable non-space ASCII. `socket.gethostname()` reports whatever the kernel stores: a space, a newline, or a byte no codec encodes, which Python hands back as a surrogate. Publishing one raw makes a holder write a record it cannot read back.

Measured against 3.32.3, on a host named `host with space`:

- `SoftReadWriteLock.acquire_read` takes the slot, then the heartbeat stops on its first tick, so the next peer ages the marker out and takes the lock from a live reader.
- `SoftReadWriteLock.acquire_write` never completes. The writer cannot recognize its own claim, so it waits out the timeout, evicts, rewrites, and loops.
- `StrictSoftFileLock.acquire` raises `SoftFileLockProtocolError`, since the claim validator rejects a hostname holding whitespace.

On a host whose name is not valid UTF-8, every writer raises `UnicodeEncodeError` before the marker exists, the soft read/write state lock included. A newline forges an extra line in a protocol 1 marker, so `SoftFileLock.is_lock_held_by_us` reports `False` for a lock this process holds.

Escaping only in the soft read/write writer leaves the strict and protocol 1 formats broken, so the escape lives in `host_name()`, the one place all four writers read the hostname. `?` is illegal in a hostname, so escaping it alongside every out-of-grammar byte as `?<hex>` leaves a conventional hostname untouched and still keeps two hosts distinct, which `owner_is_stale` relies on to refuse to probe a foreign PID.

A marker written raw by an older release on such a host no longer matches the escaped name, so `owner_is_stale` reads it as foreign and declines to reclaim it, which fails closed rather than detaching a live holder.

Tests pin the escape table, and drive each lock family through its public API on four out-of-grammar hostnames. Reverting `host_name()` to the raw name fails 11 of them.

### DIFF
--- a/docs/changelog/709.bugfix.rst
+++ b/docs/changelog/709.bugfix.rst
@@ -1,0 +1,4 @@
+``SoftReadWriteLock`` and ``AsyncSoftReadWriteLock`` now normalise the hostname written into a marker so a host whose
+``socket.gethostname()`` carries a space or a non-ASCII byte no longer publishes a marker its own heartbeat parses as
+malformed. Previously such a marker stopped the heartbeat on its first refresh and let a peer evict a still-held lock,
+and a non-ASCII hostname failed the acquire outright with ``UnicodeEncodeError``.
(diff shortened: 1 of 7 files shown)


dxbjavid:
Thanks.

PR #728, opened 2026-09-08
fix: respect ACL write access when the owner write bit is absent

jackwalkerlabs:
The lock preflight rejects any existing file without its owner-write mode bit, even when group permissions or an ACL allow the current user to write it. Consult the OS access decision before rejecting those files. Keep the writable-mode fast path so normal soft-lock contention does not mistake a concurrently removed marker for a permission error.

Reproduced with a real temporary macOS file: mode 0400 plus an ACL granting the current user write lets `os.open` succeed, but the original FileLock preflight raises PermissionError. With the fix, acquisition succeeds. Added regressions for ACL/group/other access and retained read-only/mtime tests, skipping the latter for root as real access allows root to write them.

Validation: `tox r -e 3.12,fix,type` passes: 1,290 tests, 55 capability skips, 100% line and branch coverage including tests, all repository hooks, and ty. The real ACL acquisition also passes.

Implemented and locally validated with OpenAI Codex.

Issue tracking: Found during source inspection; no matching open issue was identified for this fix.


gaborbernat (approved):

tests/test_writability.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import annotations
```
gaborbernat: Should be added to existing files, not a new file?

gaborbernat (changes requested):

PR #734, opened 2026-09-12
Fix test for musl libc

Emru1:
When packaging new version for Void Linux I've noticed that one of the tests fails on musl-based systems. Fail was due to error message, which wasn't caught by regex.


gaborbernat:
Replaced by #737

PR #739, opened 2026-09-18
fix: reject invalid poll_interval before sleep

Pitchfork-and-Torch:
## Summary
`FileLock` / `SoftFileLock` (and async variants) accepted negative, NaN, or non-numeric `poll_interval` values and only failed later inside `time.sleep` / `asyncio.sleep` with a cryptic error.

## Fix
Validate with `_resolve_poll_interval` at construction, on the property setter, and on `acquire()` / async `acquire()`, matching the existing SoftReadWriteLock timing checks (finite, non-negative).

## Test plan
- [x] `tests/test_poll_interval_validation.py` (80 passed)


tests/test_poll_interval_validation.py
```diff
@@ -0,0 +1,141 @@
+from __future__ import annotations
```
gaborbernat: No new test file, extend exisitng, and use paramterized test to minimize test code duplicaiton.

gaborbernat (changes requested):

